PR #22, opened 2023-02-08
added keys because of same widgets

mmattano:
I realized that there are issues with the keys of the user input fields. The web interface is currently not working perfectly on this branch

### DIFF
--- a/proteobench/modules/dda_quant/__init__.py
+++ b/proteobench/modules/dda_quant/__init__.py
@@ -18,7 +18,7 @@ def get_quant(
         
     replicate_quant_list = {}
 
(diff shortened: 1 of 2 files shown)


vedran-kasalica (changes requested):
The main branch doesn't have duplicate widgets currently, I think that there was a mismatch between yours and the upstream main (there is a conflict in the pull request). We can check with the group whether the duplicates are needed, just to make sure.
I also added your formatting changes to the refactoring branch manually, before I create a pull request for it (to cover that part of the pull request).

mmattano:
Sounds good! Feel free to close this PR if the duplicates are not needed

PR #57, opened 2023-02-10
:sparkles: Add black code formatting

enryH:
- currently fails
- code should be once reformatted for the entire project

Currently notebooks are not formatted

```cmd
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
```

We could enforce it using `jupyter: true`, see [the documentation](https://black.readthedocs.io/en/stable/integrations/github_actions.html)

```
(...)
- uses: psf/black@stable
  with:
    jupyter: true
```

### DIFF
--- a/.github/workflows/black-formatting.yml
+++ b/.github/workflows/black-formatting.yml
@@ -1,0 +1,10 @@
+name: Black-Code-Formatting
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


vedran-kasalica:
The lint tests are unfortunately failing, should that be fixed before the merge?

enryH:
Yes exactly I would say we need to best freeze all code and add the formatting requirement. but as black is deterministic we can also do it only once here...

PR #62, opened 2023-03-24
Caroline1

carolinelennartsson:


test/test_module_dda_quant.py
```diff
@@ -112,7 +120,25 @@ def test_MaxQuant_file(self):
...
             )
-
+            
+            
+            
+            
+class TestPlot(unittest.TestCase): 
+    """Test if the plots return a figure."""
+    
+    with open("../ProteoBench/results.json") as file:
```
vedran-kasalica, line 130: This code should be part of a method (whether the test below or a new method, in case it would be reused). I am not sure whether we should use the local `results.json` files for this test (I am not sure if this file should be even part of this repo). In case this local file is broken the test would fail. I would suggest using the actual file `https://github.com/Proteobench/Results_Module2_quant_DDA/blob/main/results.json` the raw file is available at https://raw.githubusercontent.com/Proteobench/Results_Module2_quant_DDA/main/results.json

webinterface/pages/DDA_Quant.py
```diff
@@ -90,14 +90,13 @@ def _main_page(self):
 
             with st.expander("Additional parameters"):
 
-                with open("webinterface/configuration/dda_quant.json") as file:
+                with open("../webinterface/configuration/dda_quant.json") as file:
```
vedran-kasalica, line 93: I could not test this at the moment to confirm that the change was needed, but I trust that it was justifiable.

test/test_module_dda_quant.py
```diff
@@ -112,7 +120,25 @@ def test_MaxQuant_file(self):
...
+        config = json.load(file)  
+        df = DataFrame(data=config) 
+    
+    #def test_plot_bench(self):
```
vedran-kasalica, line 134: I would remove the commented code.

vedran-kasalica (changes requested):
Most of the changes look good. I left a few minor comments. The only one that needs to be addressed is the file used for testing (`results.json`). It should probably use the file from the `Results_Module2_quant_DDA` repo in order to ensure that the tests always run correctly.

vedran-kasalica (approved):

PR #66, opened 2023-04-21
Revert streamlit cache change

scaramonche:
Hey, could you please check that? 
I think I had some weird version when I got the deprected warning. When I tested it now again (after restarting) the change I had made broke everything. So, I am not sure the st.cache_data made sense...


vedran-kasalica:
Sure, I will have a look at it

vedran-kasalica (approved):
I agree. I am not a streamlit expert, but it looks like the change should be reverted. For me locally only `cache_data` causes the error, while `cache` works fine.

vedran-kasalica:
For some reason, the tests on the main branch are failing now, which is weird, considering that this branch was not failing.

PR #67, opened 2023-06-14
Added Proline and Custom formats, as well as a first documentation

veitveit:
The following has been done:
a) Added Proline input. Works at least with the Proline output in WOMBAT. Some issues might arise from naming the quantititave columns 
b) Added a Custome tab-delimited input format
c) Added more help in the web interface explaining the intput formats


proteobench/modules/dda_quant/module.py
```diff
@@ -29,9 +30,7 @@ def generate_intermediate(
         """Take the generic format of data search output and convert it to get the quantification data (a tuple, the quantification measure and the reliability of it)."""
 
         # Summarize values of the same peptide using mean
-        quant_raw_df = filtered_df.groupby(["peptidoform", "Raw file"]).mean()[
-            "Intensity"
-        ]
+        quant_raw_df = filtered_df.groupby(["peptidoform", "Raw file"]).Intensity.mean()
```
enryH, line 33: Selects the correct numeric column as the previous behaviour of discarding non-numeric columns was depreciated in `pandas=2`

RobbinBouwmeester (changes requested):
- [ ] Could you remove the /Data/ folder? It is not clear why this is included
- [ ] Could you revert changes in the Test_workflow.ipynb

veitveit:
Requested changes should be done.

Now this also contains a first documentation in Sphinx (docs folder). I assume we still need to automate generating it in the CI.

Help is appreciated :-)

veitveit:
Now also with a github actions workflow for building the docs. No idea whether this works though

proteobench/modules/template/plot.py
```diff
@@ -6,25 +6,17 @@
...
+        fig = go.Figure()
 
-def plot_bench2(result_df):
-    """Plot results with Plotly Express."""
-    # TODO create (plotly) figure object
-    fig = go.Figure()
-
-    return fig
-
-# ? Should we define two function to implement instead of an ABC?
```
enryH, line 24: So you have chosen to not use an ABC, but a class after all? The class has no state and the module already groups the two functions. I would just use plain functions here.

enryH:
The module level documentation is really great! However, the main config for Sphinx (`conf.py`) was updated (I guess overwritten?). Was this intended? I saw Ralf is updating the readthedocs integration, so maybe we should coordinate what theme we want.

veitveit:
Wasn't the readthedocs moved to https://github.com/Proteobench/proteobench.github.io? Otherwise, yes, I overwrote it.

proteobench/modules/template/plot.py
enryH: I would just provide empty functions raising `NotImplementedError`s for the template. I'll revert the changes and make a commit.

enryH:
> Wasn't the readthedocs moved to https://github.com/Proteobench/proteobench.github.io? Otherwise, yes, I overwrote it.

@RalfG I guess it's up to you to decide:)

RobbinBouwmeester (approved):

RalfG:
> > Wasn't the readthedocs moved to https://github.com/Proteobench/proteobench.github.io? Otherwise, yes, I overwrote it.
> 
> @RalfG I guess it's up to you to decide:)

The idea back at the Copenhagen hackathon was to have dev-oriented, Sphinx-build Python API documentation on https://proteobench.readthedocs.io, and to have the main user-oriented landing page and documentation on https://proteobench.github.io (potentially forwarded to a custom domain at https://protebench.org?).

For the former I had started Sphinx docs in this repo, which Veit has partially overwritten (but that is totally fine). For the latter, I had started a Quarto-based static website on https://github.com/Proteobench/proteobench.github.io.

RalfG:
The only thing that is not entirely clear to me, @veitveit, is that you have added a GitHub action to build the docs instead of using the build system on readthedocs.org? I did not see a new build after this PR appear on https://proteobench.readthedocs.io?

veitveit:
@RalfG I added this one: https://github.com/Proteobench/ProteoBench/blob/main/.github/workflows/build_docs.yml
But did not check whether anything is triggered.

RalfG:
I saw the workflow, but it seems that `readthedocs/github-action@master` is not an existing action. With #72, docs now automatically build on the readthedocs.org servers.

PR #69, opened 2023-06-20
55 add all abstract methods to the interfaces

enryH:
Start separating interfaces.


proteobench/modules/dda_quant/parse_settings.py
```diff
@@ -39,3 +44,7 @@ def __init__(self, input_format:str):
         self.min_count_multispec = parse_settings["general"]["min_count_multispec"]
         self.species_expected_ratio = parse_settings["species_expected_ratio"]
     
+
+def parse_settings(input_format:str) -> Settings:
```
enryH, line 48: Instead of having a class to parse the setting, we could have a `Settings` dataclass instance returned by simple function. Then the type is defined. See `interfaces.py`

proteobench/modules/dda_quant/plot.py
```diff
@@ -1,12 +1,14 @@
 import numpy as np
+import pandas as pd
 import plotly.figure_factory as ff
 import plotly.graph_objects as go
 import streamlit as st
 from streamlit_plotly_events import plotly_events
 
 
+# ! This class does not use any instance attributes.
 class PlotDataPoint:
```
enryH, line 10: The PlotDataPoint class has no instance attributes. So two simple function returning the empty figures as currently done in the template should be fine

enryH:
The test's fail for `pandas=2.0` (looks like the grouping creates DataFrames with non-numeric columns)

Issue #65, opened 2023-04-21
Formatting of results.json (DDA_quant) is not consistent and sometimes not a valid JSON


vedran-kasalica:
The formatting of the 2 files is not the same:
https://github.com/Proteobench/Results_Module2_quant_DDA/blob/main/results.json (used in [main](https://github.com/Proteobench/ProteoBench) branch)
https://github.com/Proteobot/Results_Module2_quant_DDA/blob/main/results.json (used in [testing](https://github.com/Proteobench/ProteoBench/tree/json_format_testing) branch)

We have 3 possible approaches:
1. "**[Proteobench](https://github.com/Proteobench/Results_Module2_quant_DDA/blob/main/results.json)**" version uses fixed IDs (0,1,2,3,..) for each datapoint. The benefit is that it allows the identification of datapoint, however, it makes the structure hard to manage (e.g., if we delete a datapoint an ID would be missing). 
2. The "**[Proteobot](https://github.com/Proteobench/ProteoBench/tree/json_format_testing)**" version is JSON file generated by pandas, which is not a valid JSON object (such as `{"A":"1"}{"A":"2"}`), which makes it difficult to process the file with off-the-shelf tools (e.g., JSON formatters/viewers). 
3. An alternative would be to have an equivalent to **Proteobot** which is a well-formatted JSON. For example, instead of having a JSON `{"A":"1"}{"A":"2"}` we could have it formatted as an array of objects `[{"A":"1"},{"A":"2"}]`

enryH:
Regarding point 2. I have some custom code to create valid JSONs from Pandas... If I remember correctly they per default just export the numpy type.

RobbinBouwmeester:
I have had a look. I think this will fix the issue:

`all_datapoints.to_json(
        f,
        lines=True,
        orient="records"
    )`

See:

[https://github.com/Proteobot/Results_Module2_quant_DDA/pull/53](https://github.com/Proteobot/Results_Module2_quant_DDA/pull/53)

Is now on this branch:

[https://github.com/Proteobench/ProteoBench/tree/fix_json](https://github.com/Proteobench/ProteoBench/tree/fix_json)

RobbinBouwmeester:
As @vedran-kasalica mentioned I think we should now format it to an array.

RobbinBouwmeester:
Ok, I think we can address all issues with the following code:

`all_datapoints.to_json(
        f,
        orient="records",
        indent=2
    )`

See PR:

[...]

PR #70, opened 2023-07-09
Fix json

RobbinBouwmeester:
Fairly straightforward fix of writing valid JSON.

Downstream testing after accepting PR and merging this into Proteobench:

[https://github.com/Proteobot/Results_Module2_quant_DDA/pull/54/commits/3df7832d9674d3caca8d0bcf1373ba444aceb6c4](https://github.com/Proteobot/Results_Module2_quant_DDA/pull/54/commits/3df7832d9674d3caca8d0bcf1373ba444aceb6c4)

Also see: #65


vedran-kasalica (changes requested):
I agree with the changes, they solve Issue #65. I would recommend using the same on line 23 of the same file (as part of the **write_json_local_development** method).

Issue #73, opened 2023-09-15
add information on the interface of how/where to find the fasta and how to run the search, what to download etc...


mlocardpaulet:
I guess that this should be written down in GitHub/ProteoBench/webinterface/pages/DDA_Quant.py, or we should put a link to a detailed document.

mlocardpaulet:
All the information is here: 
https://docs.google.com/document/d/1cnG5q-muU8C07NYFHVtN5e8GXNByPQn4ceu6H1497zw/edit

PR #74, opened 2023-09-16
Update DDA_Quant.py

mlocardpaulet:
add information regarding DDA quan module in streamlit interface.
So... I just added some text in the page of the DDA quant module. I thought that this would be a safe start. Not touched anything else.


mlocardpaulet:
by the way, associated to issue #73

RobbinBouwmeester (changes requested):
Use triple quotes for textblock instead of escaping new line

RobbinBouwmeester (approved):

PR #91, opened 2023-09-25
Ions benchmark

RobbinBouwmeester:


mlocardpaulet:
do we want to keep the column header "peptidoform" if it is not peptidorforms anymore but ions?

RobbinBouwmeester:
Yes, agreed, would require quite a bit of work. Will fix this later.

mlocardpaulet (approved):
Approved. We just need to make sure that we change the header "peptidoforms" to "quantified_ions" in the future.

Issue #58, opened 2023-02-10
Implement support for uploading (and storing) workflow configuration files


RalfG:
Next to structured and unstructured metadata, we would like to support the upload of configuration files (e.g., mqpar.xml) to be stored alongside the intermediate files and the metadata.

enryH:
I had to parse a lot of parameter files for a submission into a table format. It is not yet a standalone module, but the functionality should be fairly general:

[Link to script to parse maxquant parameter files](https://github.com/RasmussenLab/pimms/blob/d9d6f1a3a12770bf6866ca78bef7ff3f1637167e/project/erda_05_parse_paramter_files.py)

enryH:
Aim for now:
- populate automatically as many parameters as possible in the UI based on parameter files exported by search engines
  - e.g. based on `mqpar.xml` of MaxQuant
- in case the parameter file cannot be parsed, describe the problem, but continue running (manuell editing)

mlocardpaulet:
I am making a table with the list of parameters that we want to keep, and their description. 
https://docs.google.com/spreadsheets/d/1Os3By8LYuVGcfH65X2-9tSjJnmtH-nL03tPYLtPOj-c/edit#gid=0

enryH:
If you look at e.g. json and xml files in the `test/params` folder - https://github.com/Proteobench/ProteoBench/tree/parse_settings/test/params - an example mapping could be


our parameter name |  MaxQuant |
--- | ---
MS2 mass tolerance |  '["msmsParamsArray"][0]["MatchTolerance"]'
minimum peptide length | '["minPepLen"]'

mlocardpaulet:
We noticed lack of consistency in the msfragger.params.toml file between the upper/lower tolerance bounds and the mass tolerance: 
```
precursor_mass_lower = -10			# Lower bound of the precursor mass window.
precursor_mass_upper = 10			# Upper bound of the precursor mass window.
precursor_mass_units = 1			# Precursor mass tolerance units (0 for Da, 1 for ppm).
precursor_true_tolerance = 20			# True precursor mass tolerance (window is +/- this value).
```
We need to contact the developers to ask how to deal with it.

mlocardpaulet:
precursor_mass_lower and precursor_mass_upper is defined for some pre-searching step. We'll ignore them

mlocardpaulet:
For FragPipe we need to contact the developers to ask where to find the parameters. We'll send an email when we have a define set of parameters that we need.

[...]

PR #101, opened 2023-09-25
Parse settings

enryH:
Aim: Parse parameter files of tools and map these to selected parameters for a given task for tool comparison 

- Started with `mqpar.xml` files.
- Mappings of keys between tool parameter dictionaries and selected parameters needs to be crafted per task

Issue: #58


proteobench/io/params/__init__.py
```diff
@@ -2,29 +2,30 @@
...
+    MBR: Optional[str] = None  # enable_match_between_runs
+    precursor_tol: Optional[
+        str
+    ] = None  # precursor_mass_tolerance, value and unit not separated
     precursor_tol_unit: Optional[str] = None
-    fragment_tol: Optional[str] = None
+    fragment_tol: Optional[
+        str
+    ] = None  # fragment_mass_tolerance, value and unit not separated
     fragment_tol_unit: Optional[str] = None
```
mlocardpaulet, line 23: so we should remove this one if the value is in fragment_mass_tolerance, right?

proteobench/io/params/__init__.py
```diff
@@ -2,29 +2,30 @@
...
-    MBR: Optional[str] = None
-    precursor_tol: Optional[str] = None
+    fdr_psm: Optional[str] = None  # ident_fdr_psm
+    fdr_peptide: Optional[str] = None  # ident_fdr_peptide
+    fdr_protein: Optional[str] = None  # ident_fdr_protein
+    MBR: Optional[str] = None  # enable_match_between_runs
+    precursor_tol: Optional[
+        str
+    ] = None  # precursor_mass_tolerance, value and unit not separated
     precursor_tol_unit: Optional[str] = None
```
mlocardpaulet, line 19: so we should remove this one if the value is in precursor_mass_tolerance, right?

proteobench/io/params/__init__.py
```diff
@@ -2,29 +2,30 @@
...
-
-# params = ProteoBenchParameters()
-# params.search_engine = "MaxQuant"
-# params
-# %%
+    enzyme_name: Optional[str] = None  # enzyme
+    missed_cleavages: Optional[str] = None  # allowed_miscleavages
+    min_pep_length: Optional[str] = None  # min_peptide_length
+    max_pep_length: Optional[str] = None  # max_peptide_length
+    fixed_modifications: Optional[str] = None  # fixex_mods
```
mlocardpaulet, line 28: should be 'fixed_mods'

proteobench/io/params/__init__.py
```diff
@@ -2,29 +2,30 @@
...
+    variable_modifications: Optional[str] = None  # variable_mods
+    max_num_modifications: Optional[str] = None  # max_mods
+    precursor_charge: Optional[str] = None  # min_precursor_charge
```
mlocardpaulet, line 31: if we don't enter all the possible precursor charges, then we should have min_ and max_precursor_charge

proteobench/io/params/__init__.py
mlocardpaulet, line 1: There are two parameters missing, and I did not find the correspondance in the SDRF doc. But maybe I missed it:
'software_name' for the software tool (that combines a search engine and some algorithms for quantification)
'search_engine_version'

enryH: yes, me neither. I guess this is something SDRF does intend to have.

proteobench/io/params/__init__.py
enryH: `precursor_tol` and `precursor_tol_unit` will become `precursor_mass_tolerance` - yes

proteobench/io/params/__init__.py
mlocardpaulet: I am not sure. I guess that if this happens, we'll have a missing value there.
What I meant was more that in cases like in MQ, where they give all the charges that they consider, we need to get the min. and max. No?

proteobench/io/params/__init__.py
mlocardpaulet: Could you check with Veit? Or anybody else? I think that we don't take any risk any way if we name them this way, No?

enryH: no, we can decide the name for parameters specific to our needs. SDRF is independent of software tools, so it would not make sense to have this kind of information there.

proteobench/io/params/__init__.py
enryH: MQ specifies a range of min. and max. charge. We only have a problem if a tools would e.g. allow peptides with charges 2,3 and 5, but excludes charges with 4. But I guess this is quite uncommon?

enryH: In that case yes, as we should not diverge from SDRF-extended.

enryH:
One more question regarding `fragement_mass_tolerance`. Which fractionation technique was used? MQ seems to save several settings for different methods.

And it also has two version in our 3 xml files

1.6.0.0 and higher.
https://github.com/Proteobench/ProteoBench/blob/682825d27f5fb85692e6f2884b40d79334234920/test/params/mqpar_MQ1.6.3.3_MBR.xml#L384C1-L465C31

And before:
https://github.com/Proteobench/ProteoBench/blob/682825d27f5fb85692e6f2884b40d79334234920/test/params/mqpar1.5.3.30_MBR.xml#L220C1-L249C22

mlocardpaulet:
> One more question regarding `fragement_mass_tolerance`. Which fractionation technique was used? MQ seems to save several settings for different methods.
> 
> And it also has two version in our 3 xml files
> 
> 1.6.0.0 and higher. https://github.com/Proteobench/ProteoBench/blob/682825d27f5fb85692e6f2884b40d79334234920/test/params/mqpar_MQ1.6.3.3_MBR.xml#L384C1-L465C31
> 
> And before: https://github.com/Proteobench/ProteoBench/blob/682825d27f5fb85692e6f2884b40d79334234920/test/params/mqpar1.5.3.30_MBR.xml#L220C1-L249C22

Maybe to check with Bart, but I'd say that you take the "FTMS" params. Is it what you asked?

mlocardpaulet:
> Yes, I would like to have the information double check w.r.t to all tolerances. If it is FTMS, is this also be noted somewhere in the parameter file? @brvpuyve Could you check which tolerances you would select for reporting in the two above mentioned MQ parameter files?

MQ will look for the in formation in the raw file (for the FTMS). And here we used an orbitrap so this should be it.

RobbinBouwmeester (approved):

enryH:
No you were fast:) I'll add one test later, but this is only to keep track of changes.

PR #103, opened 2023-09-25
Add additional search engine, some fixes

lazear:
Hi all, 

I stumbled across this - super cool project. I added support for the [Sage search engine](https://github.com/lazear/sage), since I happened to already have PXD028735 downloaded. 

I've also added some misc. fixes - happy to cherrypick out any you don't want. Also, if you're not ready to add in additional tools, I can hold off on the PR.
- I think the expected ratios are off for all of the tools
- I fixed the filenames to use the DDA files, not the DIA (AIF) files
- Changed deprecated plotly function to new plotly express histogram (also, old one was not properly displaying densities)
- Added tests for Sage input

Everything seems to work on all of the existing test datasets. I did notice that for some reason, the overall metrics are not being calculated correctly for Sage and displayed on the final scatterplot that compares tools (there are NaN values maybe?) despite the "Download calculated ratios" file looking OK.

### DIFF
--- a/proteobench/modules/dda_quant/parse_settings.py
+++ b/proteobench/modules/dda_quant/parse_settings.py
@@ -32,7 +32,7 @@ INPUT_FORMATS = ("MaxQuant",
                 "Sage",
                 "Custom")
 
-LOCAL_DEVELOPMENT = True
+LOCAL_DEVELOPMENT = False
 
 # For local development change below to the json and path, if you do not want to download it from github
 DDA_QUANT_RESULTS_PATH = "https://raw.githubusercontent.com/Proteobench/Results_Module2_quant_DDA/main/results.json" #e.g., K:/results.json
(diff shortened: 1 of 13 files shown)


proteobench/modules/dda_quant/parse_settings.py
```diff
@@ -28,9 +29,10 @@
                 "MSFragger",
                 "Proline",
                 "WOMBAT",
+                "Sage",
                 "Custom")
 
-LOCAL_DEVELOPMENT = False
+LOCAL_DEVELOPMENT = True
```
enryH, line 35: @scaramonche I guess you will have to decide about this change:)

RobbinBouwmeester (approved):
Indeed good catch with the wrong ratios!

proteobench/modules/dda_quant/parse_settings.py
enryH, line 35:
```suggestion
LOCAL_DEVELOPMENT = False
```

Issue #102, opened 2023-09-25
add logos


mlocardpaulet:
we need to add the logo of EuBiC, EuPA and funders (DDSA) as well as participants:
https://drive.google.com/drive/folders/1CCSVZrZmMSeTVWhbSNEZ5lvAoddABKj9

PR #117, opened 2023-09-27
Add logos - issue #102

mlocardpaulet:


webinterface/_base.py
```diff
@@ -35,15 +35,24 @@ def _preface(self):
             """,
             unsafe_allow_html=True,
         )
+        st.image("https://github.com/Proteobench/ProteoBench/blob/Add-logos/webinterface/logos/logo_participants/logos_all_20230926.png?raw=true")
```
enryH, line 38: I guess you discussed this, but what is the reason to use the github url? Wouldn't the application be shipped with the images and the relative path be enough?

webinterface/logos/logo_proteobench/source material/MuseoModerno Regular.ttf
enryH, line 1: filepath with space? does ttf format work?

webinterface/_base.py
mlocardpaulet: We did not discuss this. I just did it this way for lack of knowing better I think. 
I am totally open to suggestions :)

webinterface/pages/DDA_Quant.py
```diff
@@ -110,12 +110,12 @@ def _main_page(self):
...
                     Q-Exactive H-FX (ThermoScientific). They can be downloaded from the
-                    proteomeXchange repository PXD028735. You can download them here:
-                    [LFQ_Orbitrap_DDA_Condition_A_Sample_Alpha_01.raw](https://ftp.pride.ebi.ac.uk/pride/data/archive/2022/02/PXD028735/LFQ_Orbitrap_DDA_Condition_A_Sample_Alpha_01.raw)
-                    [LFQ_Orbitrap_DDA_Condition_A_Sample_Alpha_02.raw](https://ftp.pride.ebi.ac.uk/pride/data/archive/2022/02/PXD028735/LFQ_Orbitrap_DDA_Condition_A_Sample_Alpha_02.raw)
-                    [LFQ_Orbitrap_DDA_Condition_A_Sample_Alpha_03.raw](https://ftp.pride.ebi.ac.uk/pride/data/archive/2022/02/PXD028735/LFQ_Orbitrap_DDA_Condition_A_Sample_Alpha_03.raw)
-                    [LFQ_Orbitrap_DDA_Condition_B_Sample_Alpha_01.raw](https://ftp.pride.ebi.ac.uk/pride/data/archive/2022/02/PXD028735/LFQ_Orbitrap_DDA_Condition_B_Sample_Alpha_01.raw)
-                    [LFQ_Orbitrap_DDA_Condition_B_Sample_Alpha_02.raw](https://ftp.pride.ebi.ac.uk/pride/data/archive/2022/02/PXD028735/LFQ_Orbitrap_DDA_Condition_B_Sample_Alpha_02.raw)
+                    proteomeXchange repository PXD028735. You can download them here:
```
enryH, line 113:
```suggestion
                    proteomeXchange repository PXD028735. You can download them here:
```

enryH: trailing which spaces

enryH (changes requested):
at least the trailing white spaces should be removed and the `ttf` file could be renamed:)

webinterface/_base.py
enryH, line 38:
```suggestion
        st.image("webinterface/logos/logo_participants/logos_all_20230926.png")
```

Depending what the root folder the application should be started from, I would suspect that this is works as well. 

I just saw now: you will need to change the link once it is merged as it points to the `Add-logos` branch. So try to make them all relative paths:)

enryH, line 38:
```suggestion
        st.image("webinterface/logos/logo_participants/logos_all_20230926.png")
```

Depending what the root folder the application should be started from, I would suspect that this is works as well. 

I just saw now: you will need to change the link once it is merged as it points to the `Add-logos` branch. So try to make them all relative paths:)

enryH:
image

What should be above the DDSA logo?

mlocardpaulet:
> in principle only the main logo is needed for the application. So we would not need to provide all of the single logos in each copy of the repository.

Well, our main logos are ProteoBench, EuBIC-MS and EuPA. For the moment, these should be visible when people use the platform and when we make demos. So I think that this is good to have them on the side of the Module.

mlocardpaulet:
> image
> 
> What should be above the DDSA logo?

The ProteoBench logo

enryH:
image

So we need the for 4-5 logs in the side bar and the composite logo?

mlocardpaulet:
> image
> 
> So we need the for 4-5 logs in the side bar and the composite logo?

Oups! I really did not think of dark theme... Is it OK for you if I create a single image with white background for the side-bar logos? Maybe I will remove the DDSA logo and keep only the ProteoBench, EuBIC-MS and EuPA.

mlocardpaulet:
@enryH : I have combined the sidebar logos together, and it now has a white background. It is definitely not ideal. All these logos (main page and module page) could look a lot better. But it is not a priority. The priority is to have the logos present somewhere when we do demos.

RobbinBouwmeester:
@mlocardpaulet, so this can ben merged for now?

mlocardpaulet:
I'd say yes.

RobbinBouwmeester (approved):

PR #123, opened 2023-11-06
Add information in the documentation

mlocardpaulet:
I copied/pasted what we had in the drive in the documentation. I focused on the entry page of the doc., and the module life cycle and glossary. It is a start. I hope that I did not mess up anything.


wolski (approved):
Looks fine. Thanks

PR #148, opened 2023-12-01
add Proline input - parsing

mlocardpaulet:
I have:
- replaced the "replicate_mapper" with "condition_mapper" everywhere
- removed Wombat from the input options of the DDA quant module since it is precursor ions, and there is not output for this from Wombat. I kept the associated files though. So that we have them when we copy the script for the peptidoform module
- tried to parse the Proline input: before, we used as input a txt file that contained only the ion quantification. Now we use the excel file as input. This way, we also have the tabs with the parameters. 
**BUT**: I did not manage to parse the modifications. I don't really understand how the `Modified sequence` is generated and how the toml `[modifications_parser]` should look like. 
Maybe @RobbinBouwmeester could give me pointers?

### DIFF
--- a/docs/modules/3-DDA-Quantification-ion-level.md
+++ b/docs/modules/3-DDA-Quantification-ion-level.md
@@ -10,7 +10,7 @@ Please refer to the original publication for the full description of sample prep
 
 ## Metric calculation
 
-For each precursor ion (modified sequence + charge), we calculate the average signal per condition ("0" are replaced by NAs and missing values are ignored). The total number of unique precursor ions is reported on the *y*-axis, and the weighted sum of the mean absolute error from the expected ratio is reported on the *x*-axis. Precursors matched to contaminant sequences and/or to multiple species are excluded for error calculation.
+For each precursor ion (modified sequence + charge), we calculate the sum of signal per raw file and condition. Then, we calculate the average signal per condition ("0" are replaced by NAs and missing values are ignored). The total number of unique precursor ions is reported on the *y*-axis, and the weighted sum of the mean absolute error from the expected ratio is reported on the *x*-axis. Precursors matched to contaminant sequences and/or to multiple species are excluded for error calculation.
 
 ## How to use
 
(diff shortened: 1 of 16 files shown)


RobbinBouwmeester:
Does wombat still follow the same formatting as before (e.g.; "EFHQAGKPIGLC[Carbamidomethyl]C[Carbamidomethyl]") that should be good without parsing. If you want no parsing you can completely leave out the modifications_parser section in the .toml.

mlocardpaulet:
I was not clear, sorry. 
1. For WOMBAT: it cannot be implemented in the DDA-quant module for ions because the outputs do not report the ion quantification. So we just ignore this tool for the ion-level module, it will be implemented for the peptidoform module.
2. the modification parsing is for Proline output. In Proline, we have the stripped sequence and the modifications in two different fields:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/locard/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/locard/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


sequence | modifications
-- | --
LMSAISK | Oxidation (M2)
CLATHGR | Gln->pyro-Glu (Any N-term); Carbamidomethyl (C1)



</body>

</html>

And I don't really understand how to make the correct toml file for generating the modified sequence. 

Is this clearer?

RobbinBouwmeester:
Do you have an example of the proline file? We will need to add it to the test routine anyway.

mlocardpaulet:
> > Do you have an example of the proline file? We will need to add it to the test routine anyway.
> 
> Yep, let me find it and put it with the other test files.

Done, the file is here: https://github.com/Proteobench/ProteoBench/tree/main/test/data

RobbinBouwmeester:
I see what you mean right now, this is not yet supported, should be easily fixed, will look at this ASAP.

enryH (approved):
the renaming and switching off of WOMBAT looks good.

PR #149, opened 2023-12-01
Add plot before submitting data

RobbinBouwmeester:
Add plot before submitting data


mlocardpaulet (approved):
Looks good, 
the only little thing that could be improved is that it takes a while for the plot to appear. There is the "running" pictogram on the top right, but maybe we should think of a more visible message that would be printed. This would go on the bottom of the to-do list ;)

Issue #150, opened 2023-12-01
always use light mode


mlocardpaulet:
I think that the plot is difficult to see in dark mode. Also, it makes our life more complicated for the logos. So I would suggest to force light mode.

scaramonche:
As far as I could see one needs to create a `config.toml` file in the `webinterface/.streamlit/` directory and add the following setting to it:
```
[theme]
base="light"
```
@RobbinBouwmeester, I assume this could be done on the deployment server and I am not sure if it needs to be in the github repo of ProteoBench. What do you think?

PR #152, opened 2023-12-01
Add theme to be light by default

scaramonche:
@Robbin, I hope I understood correctly that it is fine to add the config file to github. This addresses #150.


RobbinBouwmeester:
Yes!

RobbinBouwmeester (approved):

PR #158, opened 2023-12-11
Use metadata params for submission

RobbinBouwmeester:


proteobench/modules/dda_quant/module.py
```diff
@@ -370,9 +384,33 @@ def write_intermediate_raw(
...
+
+    def load_params_file(
+        self, input_file: str, input_format: str
+    ) -> ProteoBenchParameters:
+        """Method loads parameters from a metadata file depending on its format. TODO: Currently only supports MaxQuant, MSFragger, and Proline"""
+        params = ProteoBenchParameters()
+
+        if input_format == "MaxQuant":
+            params = maxquant.extract_params(input_file)
+        # elif input_format == "AlphaPept":
```
enryH, line 403: also availabe. Would you like to have a dictionary of `extract_params` functions in the module `preteobench.io.params`? Then you could avoid the long block of if/else statements and handle potential key-errros.

scaramonche:
@RobbinBouwmeester I think it is pretty much done. Tests should work once the results.json in the repo is replaced. Note that the biggest change comes from using `software_name` as the variable that contains the `input_format` (MaxQuant, etc.) instead of `search_engine` as used until now. 
I did not change the template yet, but I can also do that later.

proteobench/modules/dda_quant/module.py
```diff
@@ -343,6 +359,8 @@ def clone_pr(
         current_datapoint = temporary_datapoints.iloc[-1]
         current_datapoint["is_temporary"] = False
         for k, v in datapoint_params.__dict__.items():
+            if k == "software_name":
```
RobbinBouwmeester, line 362: This needs to be removed?

proteobench/modules/dda_quant/module.py
```diff
@@ -380,6 +398,8 @@ def write_json_local_development(self, temporary_datapoints, datapoint_params):
 
         # Update parameters based on parsed params
         for k, v in datapoint_params.__dict__.items():
+            if k == "software_name":
```
RobbinBouwmeester, line 401: This needs to be removed?

proteobench/modules/dda_quant/module.py
scaramonche: Indeed, fixed in a (hopefully) more elegant way.

scaramonche (approved):
I think we are good to go. Maybe @enryH can have a final look as an independent observer?

PR #161, opened 2023-12-15
:heavy_plus_sign: Add selected theme as docs dependency

enryH:
Try to get [failing docs running](https://readthedocs.org/projects/proteobench/builds/22852679/).

Theme dependency is missing.


enryH:
The website will only be rebuild once we merge... If there is another error it might take a while to update everything.

PR #163, opened 2023-12-15
Issue115issue40

wolski:


wolski:
@RobbinBouwmeester 

This is where we are at the moment:
The plot methods are updated (new parameter nr_observed)
all tests pass
Application works
Adding all_datapoints read from the JSON to the new data point still needs to be implemented. This is because the data frames have different columns at the moment, and I did not know how to update the JSON. (edited) 
It works because there is a check if the columns are the same, and if not, it discards the all_datapoints

https://github.com/Proteobench/ProteoBench/blob/70a47e46126904d3abbd53c9c8a65afac891e07b/proteobench/modules/dda_quant/module.py#L262C10-L262C10

RobbinBouwmeester:
@wolski 

I have tried to continue with the code from this branch, but I had some serious issues properly integrating storing results (metrics... datapoints...) to the github repo.

This new branch should supply a much more simplistic and also more maintainable solution:

#175

wolski:
superseded by pull request #175

Issue #166, opened 2023-12-15
Change "MSFragger" to "FragPipe"


mlocardpaulet:
"MSFragger" is used as a software tool name throughout the code. It is not. It is a search engine. 
So it should be corrected to "FragPipe". 
Sorry for not catching it earlier.

scaramonche:
Addressed in #168.

PR #168, opened 2023-12-15
Rename MSFragger to FragPipe (issue #166)

scaramonche:
Since the actual software is called FragPipe and MSFragger is the search engine, we should rename all of those.


mlocardpaulet:
This looks great, thanks. I just added a few modifications in the docs.

mlocardpaulet (approved):

PR #172, opened 2023-12-15
Fix issue#122

mlocardpaulet:
I just added the ID so that we see it on hovering.


mlocardpaulet:
It is just that the FDR at protein level, and some other information may not be the most interesting for this module. And I felt that the hovering box was a bit too big.

scaramonche (commented):
Do you want to also change Search engine to be called Software tool?

mlocardpaulet:
> Do you want to also change Search engine to be called Software tool?

Oh dear yes! I should. Thanks!

mlocardpaulet:
Done!

wolski (approved):

PR #175, opened 2023-12-16
Allow for less quant

RobbinBouwmeester:


wolski:
@RobbinBouwmeester 

OK. We are almost there. 
Although, we still have a problem.
See screenshot of all_datapoints, 
image

when Module().benchmarking is called from the StreamlitUI. You can see that the second row with the new datapoint is NA for most of the columns (see screenshot above).

In the unit tests (test_benchmarking) it looks good.
image

wolski:
@RobbinBouwmeester 
Is all good. The "NA" issue is solved. There were no values provided when running from the GUI. Hence the NA's. I think you can create a pull request.

PR #178, opened 2023-12-17
added median CV, Q75 and Q90 to datapoint.result

wolski:


RobbinBouwmeester (approved):

proteobench/modules/dda_quant/module.py
```diff
@@ -172,8 +172,18 @@ def get_metrics(df, min_nr_observed=1):
         df_slice = df[df["nr_observed"] >= min_nr_observed]
         weighted_sum = round(df_slice["epsilon"].abs().mean(), ndigits=3)
         nr_prec = len(df_slice)
-
-        return {min_nr_observed: {"weighted_sum": weighted_sum, "nr_prec": nr_prec}}
+        median_cv = round((df_slice["CV_A"].median() + df_slice["CV_B"].median()) / 2, ndigits=3)
```
RobbinBouwmeester, line 175: Is there a reason we already want to round here? Could it be better to round when plotting?

wolski: The old code for weighted_sum had the round( ..., digits=3). There is no reason otherwise, therefore I am removing the round.
weighted_sum should actually be also renamed to median_abs_epsilon.

RobbinBouwmeester (approved):

Issue #154, opened 2023-12-02
histogram colors are not consistent


mlocardpaulet:
when plotting the first histogram (log2-FCs for the different species), the colors do not always corespond to the same species. For example, today I have:
human = dark blue
yeast = blue
e. coli = red
for an msfragger import, and:
human = dark blue
e. coli = blue
yeast = red

It is not a big issue but it would be better to always have the same color for the same species

wolski:
Should we add the color information to the module settings toml? For instance?

```
[species_expected_ratio]
[species_expected_ratio.YEAST]
"A_vs_B" = 2.0
"color" = "red"
[species_expected_ratio.ECOLI]
"A_vs_B" = 0.25
"color" = "blue"
[species_expected_ratio.HUMAN]
"A_vs_B" = 1.0
"color" = "green"

[General]
min_count_multispec = 1

mlocardpaulet:
> Should we add the color specification to the module_settings.tool? For instance:
> 
> ```
> [species_expected_ratio]
> [species_expected_ratio.YEAST]
> "A_vs_B" = 2.0
> "color" = "red"
> [species_expected_ratio.ECOLI]
> "A_vs_B" = 0.25
> "color" = "blue"
> [species_expected_ratio.HUMAN]
> "A_vs_B" = 1.0
> "color" = "green"
> 
> [General]
> min_count_multispec = 1
> ```

I think that this is a good idea :)

wolski:
closed by pull request 180

PR #180, opened 2023-12-18
adresses issue #154

wolski:
histogram colors are specified in toml and shown in histogram. See issue #154 for details.
Also added vertical lines to show expected fold changes.
On this occasion noticed that the ratios in the toml are not log2 ratios so need to be transformed this pull request fixes this bug in the compute_epsilon function.


mlocardpaulet (approved):
It works well on my side, the only thing that I notice is that the plot does not appear. The user has to move the slidebar with the min. num. of measured values and then it appears. 
But I think that it is not an issue with this PR, so I approve this one and we can make another issue.

PR #224, opened 2024-01-27
Fix check name/hash

### LINKED ISSUE #209: issue when public submission of the same data twice
I uploaded twice the same data for a test, and ProteoBench catches that the data is already uploade, BUT:
1. the error message reports an identifier:
"The run you want to submit has been previously submitted under the identifier: {'e789cf86011a35dd8bbac93e2e2e965cd9fc4b7b'}"
but it is not the unique ID that is reported in the figure and at upload (such as "MaxQuant__20240108_110441"), so it is not possible for the user to then find the points he/she is interested in. 
2. there is a mention:
"Follow your submission approval here", but the link is broken since there is no PR to look for.

image

RobbinBouwmeester:
@mlocardpaulet this should fix #209, can you check?

### DIFF
--- a/proteobench/modules/dda_quant/module.py
+++ b/proteobench/modules/dda_quant/module.py
@@ -323,10 +323,11 @@ class Module(ModuleInterface):
         overlap = set_current_datapoint.intersection(set_all_datapoints_old)
 
         if len(overlap) > 0:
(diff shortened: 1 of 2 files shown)


mlocardpaulet:
I tested it and it seems to work (there is no pr created), but there is still an error in the error message:

image

PR #226, opened 2024-01-29
Add CV plot

wolski:
Added CV's to DDA_Quant UI. 
Once merged, I will also sync these plots with the number of missing values selected in the slider. For the moment, there is no filtering. (not that I removed it; it was only implemented for the "Mean error between conditions" plot).


RobbinBouwmeester:
When reducing width the plots are overlapping:

Screenshot 2024-01-30 140119

wolski:
@RobbinBouwmeester I fixed the issue of overlapping figures by passing use_container_width=True to plotly_chart

RobbinBouwmeester (approved):

PR #247, opened 2024-02-11
229 integration of i2masschroq results in dda quantification ion level module

mlocardpaulet:
I have made the toml, and added "i2MassChroQ" where needed (included to set up the colors of the dots, maybe we want to make this bit automatic).
I have **not work at all on tests**. I am happy to do so but may need pointers, never did testing before.
There is an error, and I think that it is connected to proforma parsing. The error message is triggered from get_proforma_bracketed (from what I understood):
image


mlocardpaulet:
here is an example of a modified sequence in the input table: `"AALQEELQLC[MOD:00397]K"`

RobbinBouwmeester (approved):

mlocardpaulet:
looks good. @wolski , we turned off some features of the plots that you added because it was becoming too slow. We can discuss this.

proteobench/modules/dda_quant/module.py
```diff
@@ -262,6 +262,9 @@ def load_input_file(self, input_csv: str, input_format: str) -> pd.DataFrame:
                 lambda x: aggregate_modification_column(x.sequence, x.modifications),
                 axis=1,
             )
+        elif input_format == "i2MassChroQ":
+            input_data_frame = pd.read_csv(input_csv, low_memory=False, sep="\t")
+            input_data_frame["proforma"] = input_data_frame["ProForma"]
```
mlocardpaulet, line 267: Hey! Robbin, I don't understand something. Line 5 of the toml, we indicate that the proforma modified sequences are in the column "ProForma". Does it make sense to have to specify it here differently here?

PR #248, opened 2024-02-11
fix custom import

mlocardpaulet:
I fixed the custom import. I'll update the docs asap. 

One thing that I noticed though @RobbinBouwmeester:
in module.py, there is this line:
`input_data_frame["proforma"] = input_data_frame["Modified sequence"]`
But isn't it redundant with the toml file line that defines the column with the modified sequence?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,7 +13,7 @@ Overview of ProteoBench
     Changelog <changelog.rst>
 
 
-ProteoBench is an open and collaborative platform for community-curated benchmarks for proteomics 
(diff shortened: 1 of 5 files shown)


mlocardpaulet:
here is a test file
[test_custom.txt](https://github.com/Proteobench/ProteoBench/files/14232782/test_custom.txt)

RobbinBouwmeester:
Regarding this bit:

`input_data_frame["proforma"] = input_data_frame["Modified sequence"]`

This is search engine/software specific parsing. By this point I believe we have not read the input toml yet. So either we need to read this beforehand or keep it as is. I would opt for the last option and make a todo/comment in the code that we might change this in the future.

PR #258, opened 2024-02-28
Peptidoform and peptide modules

RobbinBouwmeester:


wolski:
The files in io_parse_settings in dda_quant_base and dda_quant_ion are identical. They can be removed from "dda_quant_base".
Shall we create a subfolder for each module, e.g. 
test/data/DDA_ion etc?


Do we have peptidoform files in test/data for the peptideform module? Maxquant reports peptide level quantification in the peptide.txt file. But this is the peptide (stripped sequence) level.
Do we benchmark software not reporting quantification on the peptide form level with this module, e.g., by rolling up (aggregating) precursor or PSM intensities to the peptidoform level?

wolski:
When running the "Parse bench" in the Qunat Peptidoform Level - BETA -

image

Tested on 70fa3a7 and later commits

wolski:
Also, there are now three parse.py files in dda_quant_base in dda_quant_ion and dda_quant_peptidoform
There is some code duplication among those files, e.g. 
convert_to_standard_format is present in base and peptidoform, 
aggregate_modification_column in base and peptidoform, etc.

PR #266, opened 2024-03-08
update ProteoBench on server after update

julianu:
I created a script on the server, that should update to the most-recent release (tag).
Assuming this is the correct action for this, it should work.
@RobbinBouwmeester , could you please review.


RobbinBouwmeester (approved):
Looks great! I assume any safety issues are correctly handled on the server side, but from this side it looks good!

julianu:
Should be reasonably safe, yes.
Looking forward to the next release :)

PR #287, opened 2024-04-21
:art::fire: Remove msfragger .params as dependency for FragPipe

enryH:
As pointed out by Witold, FragPipe configuration file contains all necessary information. (MsFragger .params is derived from .workflow file)


enryH:
Can I have your parameter file @brvpuyve for testing. Did you already submit your dataset? otherwise I could it for testing.

brvpuyve:
Hi @enryH,

When I tried to submit the MSFragger output it was using the public server. I think my submission was not accepted yet, since I do not see any MSFragger output. 

The worfklow file: [fragpipe.zip](https://github.com/Proteobench/ProteoBench/files/15058559/fragpipe.zip)

Please let me know if you need support!

PR #292, opened 2024-04-25
Fix local storage

RobbinBouwmeester:


enryH (commented):
So you circumvent the problem but not passing the variables, but by getting them directly from the session state?

webinterface/pages/DDA_Quant_ion.py
```diff
@@ -428,6 +426,15 @@ def create_submission_elements(self) -> None:
...
                 self.variables_dda_quant.all_datapoints
             ].copy()
+        if st.session_state[self.variables_dda_quant.input_df] is not None:
+            st.session_state[self.variables_dda_quant.input_df_submission] = st.session_state[
+                self.variables_dda_quant.input_df
+            ].copy()
+        if st.session_state[self.variables_dda_quant.result_perf] is not None:
+            st.session_state[self.variables_dda_quant.result_performance_submission] = st.session_state[
+                self.variables_dda_quant.result_perf
+            ].copy()
```
enryH, line 436: Is it necessary to make a copy in memory?

RobbinBouwmeester: Yes, as the others are reset when you upload the params file.

enryH: Probably it only create a new reference without copying the underlying data? We would need to check what the current default is in pandas

enryH (approved):

Issue #291, opened 2024-04-25
:bug: QuantScores


enryH:
Then I run into an error related to QuantScores:
```
File "C:\Users\kzl465\Documents\repos\ProteoBench\webinterface\pages\DDA_Quant_ion.py", line 238, in _run_proteobench
    result_performance, all_datapoints, input_df = self.ionmodule.benchmarking(
File "C:\Users\kzl465\Documents\repos\ProteoBench\proteobench\modules\dda_quant_ion\module.py", line 42, in benchmarking
    self.precursor_name, parse_settings.species_expected_ratio(), parse_settings.species_dict()
TypeError: 'dict' object is not callable
```

https://github.com/Proteobench/ProteoBench/blob/8284e9c5f8764cbf8ea0e66e40c1bb10e28eba0b/proteobench/modules/dda_quant_ion/module.py#L40-L43

Which is probably due to an attribute and the corresponding method having the same name:

https://github.com/Proteobench/ProteoBench/blob/8284e9c5f8764cbf8ea0e66e40c1bb10e28eba0b/proteobench/io/parsing/parse_settings_ion.py#L43-L63

which is also the case for the base class: 

https://github.com/Proteobench/ProteoBench/blob/bc7233784a67f285f9d019c9bc08bcbc77661754/proteobench/score/quant/quantscores.py#L7-L12

So I guess removing the assessor method for the dicts should be the best solution? @wolski what was you idea to create these?

wolski:
@enryH How did you run into the error? I can not replicate it. Are you on branch main?


parse_settings is not a dict but a `class ParseSettings:`

and we need the assesors because ParseModificationSettings and ParseSettings need to have the same interface
but have different implementations:

```
ParseModificationSettings
def species_dict(self):
        return self.parser.species_dict

    def species_expected_ratio(self):
        return self.parser.species_expected_ratio
```

```
ParseSettings
 def species_dict(self):
        return self.species_dict

    def species_expected_ratio(self):
        return self.species_expected_ratio
```

enryH:
I ran into the error when I tried to integrate i2MassChroQ into the UI. I am not 100% sure why yet: https://github.com/Proteobench/ProteoBench/issues/229

PR: https://github.com/Proteobench/ProteoBench/pull/279

So I am working on that branch. 

[...]

PR #293, opened 2024-04-26
fixing problem reported in issue #291

wolski:


enryH (approved):
Do you know how `ParseSettingsBuilder` is related to the `ParseSettings`. Is it a factory for ParseSettings? (I have not checked the code, but the naming suggest it)

PR #295, opened 2024-05-03
Add version to docs

enryH:
- add version tag to website


mlocardpaulet:
> https://proteobench--295.org.readthedocs.build/en/295/

It would be great. It could be better to have the version visible from all pages (next to the top menu for example?), but if it is even a bit complicated, don't bother.

mlocardpaulet (approved):

PR #305, opened 2024-05-21
Combine protein columns fragpipe

RobbinBouwmeester:


mlocardpaulet:
Thanks @RobbinBouwmeester. So in the end, the FragPipe .toml is not entirely correct when saying 
```Protein = "Proteins"```
Shall I modify it? Even with just a comment?
I should also modify the documentation to explain what we do, right?

PR #307, opened 2024-05-21
Fix double brackets and C-terminal mods

RobbinBouwmeester:


mlocardpaulet (approved):
This works well on my side.

Issue #308, opened 2024-05-25
Size limit for the software output files (ions)


mlocardpaulet:
When I want to upload an alphapept `results_peptides.csv` on the sever, I have this error message because the file is too big: 
<img width="548" alt="image" src="https://github.com/Proteobench/ProteoBench/assets/8386272/06b034bb-292d-4b3b-969a-818c3d2e326f">

mlocardpaulet:
the files I have are 202 and 225 MB (2 runs with different parameters).

mlocardpaulet:
Thanks! It works nicely :)

PR #311, opened 2024-05-27
:bug: Allow local use and UploadFile use in streamlit. Fix #310

### LINKED ISSUE #310: issue with alphapept parameter parsing
**Describe the bug**
I have an error message when I upload the .yaml file from and alphapept file for public upload. Everything before this worked out well. 

**To Reproduce**
1. Parse and bench an alphapept `results_peptide.csv` output file
2. the plot loads nicely
3. Scroll down to public upload, and load the `.yaml` corresponding to the search (I also tried the `results.yaml` without success)
4. See error (screenshot below)

**Expected behavior**
the parameter files should be parsed for public uploac

**Screenshots**
image

I can provide a zip with the associated files :)

enryH:
- default is a BufferIO like object from streamlit.

Closes #310

### DIFF
--- a/proteobench/io/params/alphapept.py
+++ b/proteobench/io/params/alphapept.py
@@ -9,8 +9,13 @@ from proteobench.io.params import ProteoBenchParameters
 
 
 def extract_params(fname) -> ProteoBenchParameters:
-    with open(fname) as f:
-        record = yaml.safe_load(f)
+    try:
+        record = yaml.safe_load(fname)
+    except AttributeError:
+        print(f"Load locally: {fname}")
+        with open(fname, "rb") as f:
+            record = yaml.safe_load(f)
+
     summary = record["summary"]
     params = ProteoBenchParameters()
     params.software_name = "AlphaPept"


mlocardpaulet:
When I try locally I have the same issue as issue #308 (was corrected by @Julian). It says that the file is too big. 
image

enryH:
It's probably the local configuration issue I wrote about before. Modify the `webinterface/.streamlit/config.yaml` the following way. 

```toml
[theme]
base = "light"

[server]
enableStaticServing = true
# added: 
enableXsrfProtection = false
enableCORS = false
```

It's a security feature for local execution to increase security.

julianu:
Yes, I set the `maxUploadSize`to 2 GB on the server for now.

What are these two for?
```
enableXsrfProtection = false
enableCORS = false
```

mlocardpaulet:
Thanks Henry, it works very nicely :)
image

mlocardpaulet (approved):

mlocardpaulet:
@enryH should we change the `webinterface/.streamlit/config.toml` file in the repo?

enryH:
> @enryH should we change the `webinterface/.streamlit/config.toml` file in the repo?

This is a question for @julianu . I think we should rather document it as it's a security feature for local execution which should not be deactivate when published. So it will depend on how the server is configured now.

enryH:
> Yes, I set the `maxUploadSize`to 2 GB on the server for now.
> 
> What are these two for?
> 
> ```
> enableXsrfProtection = false
> enableCORS = false
> ```

If I do not set it, my browser does not allow to load files from the local disk. It's probably a security feature.

PR #314, opened 2024-06-04
adding gdpr notices if provided

julianu:
This addition shows links to "privacy notice" and "legal notice" on a hosted system. This is only rendered if the data is given in the `secrets.toml`.

Also, the "hosting information" is now only rendered if a specific string is provided in the `secrets.toml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
   "openpyxl",
   "numpy",
   "plotly",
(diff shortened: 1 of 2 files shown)


julianu:
Nice rendering of the links  requires streamlit to be at least version 1.31, which should be no problem.

RobbinBouwmeester (approved):

PR #315, opened 2024-06-04
:memo: harmonize version in CHANGELOG.md

enryH:


enryH:
There is additional configurations: https://github.com/anton-yurchenko/git-release?tab=readme-ov-file#configuration

PR #317, opened 2024-06-12
refactoring the sidebar and general configs

julianu:
This PR refactors the call for the sidebar, so it is only defined in one position.
Also, allows a secret-setting for website tracking / analysing.

As soon as this is merged (and a new version created), we should be able to track the traffic on our server page.

### DIFF
--- a/webinterface/pages/texts/proteobench_builder.py
+++ b/webinterface/pages/texts/proteobench_builder.py
@@ -1,0 +1,23 @@
+"""Streamlit-wide page settings and tools for ProteoBench."""
+import streamlit as st
+
+def proteobench_page_config(page_layout = "wide"):
+    """Set some ProteoBench wide page settings"""
+    st.set_page_config(
+        page_title="Proteobench",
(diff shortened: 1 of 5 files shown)


webinterface/pages/texts/proteobench_builder.py
```diff
@@ -0,0 +1,23 @@
...
+        initial_sidebar_state="expanded",
+    )
+
+def proteobench_sidebar():
+    """Format the sidebar for ProteoBench."""
+    st.sidebar.image("logos/logo_funding/main_logos_sidebar.png", width=300)
+
+    # add gdpr links if provided
+    if "gdpr_links" in st.secrets.keys():
+        st.sidebar.page_link(st.secrets["gdpr_links"]["privacy_notice_link"], label="-> privacy notice")
```
enryH, line 19: Are the secrets from the yaml file? So testing this locally won't go into this part. But I trust that you tested it?

enryH (approved):

PR #318, opened 2024-06-14
:bug: update publishing workflow

enryH:
- see if flit-scm helps, otherwise change workflow to twine...


enryH:
Change the publishing workflow to be compatible with setuptools-scm, which is based on using `importlib.metadata`.

mlocardpaulet (approved):

PR #340, opened 2024-08-09
:bug: Close 338 parsing MQ 2.6 (minPepLen)

enryH:
- MaxQuant 2.6 changed the a relevant parameter name:
- minPepLen -> minPeptideLength


mlocardpaulet (approved):
I couldn't test it locally, not sure why - I have an issue with the token of the secret.toml

PR #363, opened 2024-09-03
Rename sample files

scaramonche:
Made sure it is clear which sample file belongs to which software and created an "outdated" files dir


RobbinBouwmeester (approved):

scaramonche:
Also added i2MassChroQ to the testing of dda quant.

PR #376, opened 2024-09-04
Renamed Proline to Proline Studio

scaramonche:
As suggested by Marie, we renamed Proline to Proline Studio so it is clear which software we support as there is difference between Proline Studio and Proline Web.


scaramonche:
Hm, forgot to run the tests, looking into it!

RobbinBouwmeester (approved):

PR #392, opened 2024-09-05
Show the corrrect comment fields and information depending on whether they are private of public submissions

scaramonche:


scaramonche:
Still need to add a second text field for the submission comments...

RobbinBouwmeester (approved):

PR #395, opened 2024-09-06
Add support for FraggerDIA

rodvrees:


rodvrees:
One question that I have about this is that, by default, quantification output is a DIA-NN report.tsv file. I'm assuming the user should specify MSFragger-DIA and not DIA-NN in the web interface. However, it is possible to do quantification with Ionquant as well, which will result in an MSFragger output file. What should we do about this? Have multiple parsing options for MSFragger-DIA, stick to the default workflow, ...?

rodvrees:
#396

RobbinBouwmeester (approved):

PR #425, opened 2024-10-28
:bug: temporay file alwas opened as text object, even if it's binary

enryH:
Opening the temporay file in `r` mode opens  it as a binary file, creating something like

```
 <_io.TextIOWrapper name='/var/folders/l5/ajfaa/T/tmp6asd mode='r' encoding='UTF-8'>
```

In case of `ProlineStudio` a binary excel file is submitted. This will mean that the file cannot be read by pandas (as the format is not utf8).

- [x] need to test if other tools csv files still work

`st.file_uploader` already return a File like object, which can be used. See [docs](https://docs.streamlit.io/develop/api-reference/widgets/st.file_uploader)

### DIFF
--- a/webinterface/pages/base_pages/quant.py
+++ b/webinterface/pages/base_pages/quant.py
@@ -479,9 +479,7 @@ class QuantUIObjects:
         """Executes the benchmarking process and returns the results."""
         with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
             tmp_file.write(self.user_input["input_csv"].getbuffer())
-            # temp_file_name = tmp_file.name
-        # f = open(temp_file_name)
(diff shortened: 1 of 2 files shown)


webinterface/pages/base_pages/quant.py
```diff
@@ -480,8 +480,8 @@ def run_benchmarking_process(self):
         with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
             tmp_file.write(self.user_input["input_csv"].getbuffer())
```
enryH, line 481: this invalidates the buffer, which needs to be reset if it should be used

RobbinBouwmeester (approved):

PR #430, opened 2024-11-05
MSAID output compatibility

rodvrees:
Includes a Jupyter Notebook to prepare MSAID output: protein information is not added to precursor information, only an ID. The Jupyter Notebook adds protein information based on the protein ids from proteingroups.tsv

Also converts extracted param: 'quantification_method_DIANN' into 'quantification_method'. Was originally meant specifically for different quant methods implemented in DIA-NN, but MSAID also has a bunch of quantification options, so makes sense to generalize the param. 

Also includes a fix that allows output of AlphaDIA two-step search, where it can happen that species information from fasta gets lost


RobbinBouwmeester:
@rodvrees ready to merge?

rodvrees:
@RobbinBouwmeester Yes, I'm ready!

proteobench/datapoint/quant_datapoint.py
```diff
@@ -53,6 +53,7 @@ class Datapoint:
     intermediate_hash: str = ""
     results: dict = None
     median_abs_epsilon: int = 0
+    # TODO nr_prec doesnt get updated in final df, always 0
```
RobbinBouwmeester, line 56: Remove this comment? Is resolved?

test/params/Proteobench_default_actual_details.csv
RobbinBouwmeester, line 1: Could we give this a name that makes more sense? E.g., add MSAID?

test/params/Proteobench_default_actual_details.tsv
RobbinBouwmeester, line 1: Could we give this a name that makes more sense? E.g., add MSAID?

RobbinBouwmeester (changes requested):

test/params/Proteobench_default_actual_details.tsv
rodvrees: Done

RobbinBouwmeester (approved):

PR #439, opened 2024-11-13
Plot Median AE instead of mean, add mean to metrics

### LINKED ISSUE #436: Datapoint metric is called 'median_abs_epsilon' but is actually mean
Median_abs_epsilon is calculated here: 

https://github.com/Proteobench/ProteoBench/blob/8ed5b5ad9588b5b8b10c3b0cfd9ec284be43a59a/proteobench/datapoint/quant_datapoint.py#L124-L127

So is actually mean. Since the comment suggests that the metric should be unaffected by outliers, should we actually calculate the median instead or just rename the metric to mean_absolute_median. In the scatter plot, in which this value is plotted on the x-axis, the label correctly says it's the mean absolute error.

rodvrees:
Fixes #436 according to @mlocardpaulet's suggestion

Important: the changes made here do mean that all the datapoints currently plotted are plotted incorrectly, because metrics for these datapoints cannot be retroactively recalculated (I think)


mlocardpaulet:
This will shift the points to the left, so we need to re-upload them all. We (I) did this once. I could do it again if needed.

wolski:
@mlocardpaulet We must wait with the merge, until we set up a way to recompute all the points?

mlocardpaulet:
> @mlocardpaulet We must wait with the merge, until we set up a way to recompute all the points?

Yes. I think that this is necessary.

RobbinBouwmeester:
> > @mlocardpaulet We must wait with the merge, until we set up a way to recompute all the points?
> 
> Yes. I think that this is necessary.

Would like to get this up and going ASAP. Should be done before Christmas.

mlocardpaulet:
> > > @mlocardpaulet We must wait with the merge, until we set up a way to recompute all the points?
> > 
> > 
> > Yes. I think that this is necessary.
> 
> Would like to get this up and going ASAP. Should be done before Christmas.

In the end, the idea is to:
- have the plot with mean values showing up by default when opening the module page
- have a switch button that would allow to plot the median for all points. 

Now the mean AND median are calculated at each submission, but we do not have the median values for points that were uploaded before now. 

There is an issue opened for setting up the automatic re-running of all points already submitted (while keeping their unique ID and associated comments). Since the "main" plot will not change, maybe we do not have to wait for the issue to be fixed. The only thing is that until we are able to re-run everything there will have points missing in the plot with the median.

rodvrees:
Mentioned in original issue, but relevant here too:
https://github.com/Proteobench/ProteoBench/issues/436#issuecomment-2516515500

RobbinBouwmeester:
> > > > @mlocardpaulet We must wait with the merge, until we set up a way to recompute all the points?
> > > 
> > > 
> > > Yes. I think that this is necessary.
> > 
> > 
> > Would like to get this up and going ASAP. Should be done before Christmas.
> 
> In the end, the idea is to:
> 
> * have the plot with mean values showing up by default when opening the module page
> * have a switch button that would allow to plot the median for all points (instead of mean).
> 
> Now the mean AND median are calculated at each submission, but we do not have the median values for points that were uploaded before now.
> 
> There is an issue opened for setting up the automatic re-running of all points already submitted (while keeping their unique ID and associated comments): #470. Since the "main" plot will not change, maybe we do not have to wait for the issue to be fixed. The only thing is that until we are able to re-run everything there will have points missing in the plot with the median?

Saving of the raw data is implemented. Only thing that remains is a notebook that allows for automatic reuploading. But I think we should do that manually for now. Are you OK with that.

Automatic reupload should be done from:

https://proteobench.cubimed.rub.de/datasets/

As this is currently not supported (most saved submissions are wrong; or incomplete). Only in the latest version everything is correct.

proteobench/plotting/plot_quant.py
```diff
@@ -205,7 +205,7 @@ def plot_metric(
             width=700,
             height=700,
             xaxis=dict(
-                title="Mean absolute difference between measured and expected log2-transformed fold change",
+                title="Median absolute difference between measured and expected log2-transformed fold change",
```
RobbinBouwmeester, line 208: Shouldnt the title be changed based on what is plotted?

RobbinBouwmeester (changes requested):

proteobench/plotting/plot_quant.py
rodvrees: Yes, and that's what's currently implemented on https://github.com/Proteobench/ProteoBench/tree/mean_median_plot

I'll make a new PR from that branch, then I think this one should be closed without merging

mlocardpaulet:
> > > > > @mlocardpaulet We must wait with the merge, until we set up a way to recompute all the points?
> > > > 
> > > > 
> > > > Yes. I think that this is necessary.
> > > 
> > > 
> > > Would like to get this up and going ASAP. Should be done before Christmas.
> > 
> > 
> > In the end, the idea is to:
> > 
> > * have the plot with mean values showing up by default when opening the module page
> > * have a switch button that would allow to plot the median for all points (instead of mean).
> > 
> > Now the mean AND median are calculated at each submission, but we do not have the median values for points that were uploaded before now.
> > There is an issue opened for setting up the automatic re-running of all points already submitted (while keeping their unique ID and associated comments): #470. Since the "main" plot will not change, maybe we do not have to wait for the issue to be fixed. The only thing is that until we are able to re-run everything there will have points missing in the plot with the median?
> 
> Saving of the raw data is implemented. Only thing that remains is a notebook that allows for automatic reuploading. But I think we should do that manually for now. Are you OK with that?
> 
> Automatic reupload should be done from:
> 
> https://proteobench.cubimed.rub.de/datasets/
> 
> As this is currently not supported (most saved submissions are wrong; or incomplete). Only in the latest version everything is correct.

Yes, thanks. Manual re-upload with a Jupiter notebook is great. 
If the old submissions cannot be reloaded, we can add them again. I can make this Happen (Holda has all the MQ data, Olivier can re-upload the i2MassChroQ, not sure about alphapept but I can find out). 
We'll just need to "clean up" the server and remove all wrong submissions. 
Just let me know when we want to do that and if/how I can help.

PR #445, opened 2024-11-21
Add protein columns to alphadia and spectronaut TOMLs

rodvrees:
Because some DIA search engines parse fasta headers themselves and sometimes lose species info in output, we previously implemented a mapping function to map gene names to correct fasta header info, including species info. This way we added a new protein column to the intermediate format, which negated the need to add a specified protein column in the search engine-specific TOML (because default name was used).

As per @mlocardpaulet's suggestion, we should still include protein column name information in the tomls, to avoid confusion as to which column exactly is used. 
The easiest fix would be to just keep the functionality the same, but do the mapping within the original column instead of making a new "Proteins" column. In this way, the original column name can be specified in the .toml

Let me know if this approach is okay


mlocardpaulet:
Sounds good to me. 
Thinking about it, we could also consider doing this mapping for all the pipelines (if we want to homogenise the process).
I mean... 
We would take the protein column (as identified from the toml), and:
1. check if the species are there: 
if yes -> take this information as species
if no -> do the mapping with the mapping table (accession -> species).

I suggest that without really knowing what the code looks like so feel free to completely discard this suggestion. I think that one good thing if we do it this way is that if users set the accession parsing differently, it will still work. 
Would this make sense?
If it is too much work for little reward, please do what you suggested.

rodvrees:
We could probably generalize it a bit, e.g. add an entry to the .toml that specifies if the mapper needs to be used or not. That might indeed be cleaner.

mlocardpaulet:
> We could probably generalize it a bit, e.g. add an entry to the .toml that specifies if the mapper needs to be used or not. That might indeed be cleaner.

you don't think that this can be checked automatically? I mean... If there is no "_HUMAN" (or another species), the mapper is needed.

RobbinBouwmeester:
@rodvrees ready for review?

rodvrees:
Sure, and then I'll make a new issue/PR to generalize the mapping functionality

RobbinBouwmeester (approved):

PR #451, opened 2024-11-22
homogenise peptidoforms for fragpipe and i2masschroq

Cajac102:
Added parsing for the "MOD:x" proforma formatting of i2masschroq, changed Acetyl in FragPipe to Acetylation


RobbinBouwmeester (changes requested):
Not sure anymore what is used for the other parsing, but according to unimod the PSI-MS name is Acetyl: https://www.unimod.org/modifications_view.php?editid1=1

Not sure what we should follow, but I think following PSI-MS in unimod first, followed by interim name is probably best?

RobbinBouwmeester:
Changes introduced to fit into main: #475

PR #452, opened 2024-11-25
diaPASEF module

rodvrees:


rodvrees:
Notes: 
- Having .toml files for each module is perhaps overkill, as it implies tool output parsing has to be re-implemented (to some extent) for each module, which is not desired. However, even small changes in datasets used render a .toml useless (for example, runs in the AIF module are differently named than the diaPASEF runs). What's the best way to handle this.

RobbinBouwmeester:
I actually do not have any further comments @rodvrees, maybe ready for review? And I see no reason not to merge.

rodvrees:
I have an edge case that needs fixing, maybe you can help @RobbinBouwmeester :

When I open the web interface and do a benchmarking run in the diaPASEF module, it works as desired.
However, when I open the web interface, do a benchmarking run in the AIF module (which works), and then straight after that in the diaPASEF module (without refreshing first). The 'Results new data' tab will show the plot with the data from the AIF json, with the new run. As if the wrong json is loaded (I think it has something to do with below code, where submitted_data is already assigned from earlier and not re-assigned). I can't find it how to solve it immediately, any idea where to look?

https://github.com/Proteobench/ProteoBench/blob/2b328c23c9e4c4a3dbe1fa7a03e2fc53e28cb555/webinterface/pages/base_pages/quant.py#L256C1-L262C14

RobbinBouwmeester:
I know where this is, it is likely in the page variables. It safes the results of diaPASEF in the session state, but the same var name is used to retrieve it from the session state for AIF.

FYI, I am going to work from this branch on another branch to avoid too many merge conflicts... But the sooner we can merge the better ;)

rodvrees:
Of course, I forgot to change 1 variable name... We should be good to go now

RobbinBouwmeester (approved):

PR #469, opened 2024-12-03
Add submission comment to written output

JuliaS92:
Add a file comment.txt to the data stored, containing the submission comment.


RobbinBouwmeester:
I do wonder why we would double store this next to the already downloadable json that stores comments? 

E.g., see: https://github.com/Proteobot/Results_quant_ion_DDA/blob/main/results.json

JuliaS92:
The idea is that anything we need for recreating the same results.json should be stored outside of it. This way we can recalculate from the dataset folder, including the comments.

RobbinBouwmeester (approved):

PR #489, opened 2024-12-09
Change median or mean quant error viz depending on User's wishes

rodvrees:
Replaces old PR: #439, after discussion was held on which metric should be plotted. Changes include a selection box for which metric should be plotted. Should be tested after recalculating of already submitted points is possible.


rodvrees:
Tests fail because mean_abs_epsilon is not available in the already submitted datapoints.

PR #490, opened 2024-12-09
472 make the dda module compatible with msangel outputs

mlocardpaulet:
work in progress. 
I think that the local upload work, but I have to:
- [x] make the documentation
- [ ] parse the parameters for public upload
- [ ] add tests


mlocardpaulet:
OK, so it should work for local visualisation of results only. no parsing of the parameters yet, but I think that we can merge as is and work on the parameter parsing at a later stage.

RobbinBouwmeester (approved):

PR #491, opened 2024-12-11
477 how to handle xtandem hidden parameters that have an impact on final results

mlocardpaulet:
Hi!

It now adds ",Acetyl(N-term)" when "protein, quick acetyl	yes", and ",Pyrolidone(N-term)" when "protein, quick pyrolidone	yes" in the parameter file, if the search engine is X!Tandem. 

I also changed the test result. 

It is not clean because the modification format is not the same as the modifications automatically parsed ("15.99491@M"), but I am not sure what would correspond to "Pyrolidone N-ter" in this fomat. I have to look for it.

There was a change made to the `vscode/settings.json`, I changed it back. I think. Not sure. Sorry about that. Please let me know if I messed this bit up.

### DIFF
--- a/proteobench/io/params/i2masschroq.py
+++ b/proteobench/io/params/i2masschroq.py
@@ -44,15 +44,11 @@ def extract_params(fname: pathlib.Path) -> ProteoBenchParameters:
 
     # Add "hidden" modifications when using X!Tandem:
     var_mods = ""
-    print(var_mods)
     if params.loc["AnalysisSoftware_name"] == "X!Tandem":
-        if params.loc["Protein, quick acetyl"] == "yes":
+        if params.loc["protein, quick acetyl"] == "yes":
             var_mods = ",Acetyl(N-term)"
-        if params.loc["Protein, quick pyrolidone"] == "yes":
-            var_mods = var_mods + ",Pyrolidone(N-term)"
-
-    print(var_mods)
-    # print(var_mods + ",".join(params.loc[params.index.str.contains("residue, modification mass")].dropna()))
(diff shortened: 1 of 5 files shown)


mlocardpaulet:
I'll quickly add information in the docs to explain that we do this.

proteobench/io/params/i2masschroq.py
```diff
@@ -58,8 +66,10 @@ def extract_params(fname: pathlib.Path) -> ProteoBenchParameters:
         allowed_miscleavages=max_cleavage,
         min_peptide_length=None,  # "spectrum, minimum fragment mz"
         max_peptide_length=None,  # Not mentioned, up to 38 AA in peptides
-        fixed_mods=",".join(params.loc[params.index.str.contains("residue, modification mass")].dropna()),
-        variable_mods=",".join(params.loc[params.index.str.contains("residue, potential modification mass")].dropna()),
+        fixed_mods=",".join(params.loc[params.index.str.contains(
```
enryH, line 69: so it's really an extra.

enryH (approved):
Looks good:) So what does `15.99491@M` mean?

proteobench/io/params/i2masschroq.py
mlocardpaulet: yep. These modifications are added in the search if the "quick" mod parameter is on.

mlocardpaulet:
> Looks good:) So what does `15.99491@M` mean?

This would be a way to say "methionine oxidation" with the mass of the modification and the residue.

RobbinBouwmeester (approved):

PR #509, opened 2024-12-17
parsing of i2masschroq parameters when using sage

mlocardpaulet:
add test file for parameters with sage and rename old xtandem parameter test file


mlocardpaulet:
@enryH I renamed the parameter test file with i2masschroq using xtandem, and added the one using sage (that we now need to parse). In the process, I noticed that this file `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel.csv` (renamed`test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel_xtandem.csv`) does not seem to be used. Or did I miss something? 
If not, shall we remove it?

mlocardpaulet:
> @enryH I renamed the parameter test file with i2masschroq using xtandem, and added the one using sage (that we now need to parse). In the process, I noticed that this file `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel.csv` (renamed`test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel_xtandem.csv`) does not seem to be used. Or did I miss something? If not, shall we remove it?

Actually there is also a file `i2mq_result_parameters.tsv`.

enryH:
> @enryH I renamed the parameter test file with i2masschroq using xtandem, and added the one using sage (that we now need to parse). In the process, I noticed that this file `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel.csv` (renamed`test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel_xtandem.csv`) does not seem to be used. Or did I miss something? If not, shall we remove it?

The test fail because it is missing. the `csv` version is always the version I want to see parsed ('truth'). I'll fix the tests now.

mlocardpaulet:
more information on the parameters for running sage here: https://sage-docs.vercel.app/docs/configuration#file

mlocardpaulet:
> > @enryH I renamed the parameter test file with i2masschroq using xtandem, and added the one using sage (that we now need to parse). In the process, I noticed that this file `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel.csv` (renamed`test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel_xtandem.csv`) does not seem to be used. Or did I miss something? If not, shall we remove it?
> 
> The test fail because it is missing. the `csv` version is always the version I want to see parsed ('truth'). I'll fix the tests now.

Thanks. I get it, sorry. 
And this one? `i2mq_result_parameters.tsv` it seems alone and unused. No?

enryH:
I rename the filed to be in line with my convention (as I create the `_sel.csv` versions automatically), see [here](https://github.com/Proteobench/ProteoBench/blob/417-new-support-for-i2masschroq-using-sage-instead-of-xtandem/proteobench/io/params/i2masschroq.py#L88-L108)

I rename the file again to `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_xtandemsel_.csv`, so that the test pass:)

enryH:
> > > @enryH I renamed the parameter test file with i2masschroq using xtandem, and added the one using sage (that we now need to parse). In the process, I noticed that this file `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel.csv` (renamed`test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel_xtandem.csv`) does not seem to be used. Or did I miss something? If not, shall we remove it?
> > 
> > 
> > The test fail because it is missing. the `csv` version is always the version I want to see parsed ('truth'). I'll fix the tests now.
> 
> Thanks. I get it, sorry. And this one? `i2mq_result_parameters.tsv` it seems alone and unused. No?

 I think it's an updated version of the file I am using in the test. Should we use both or either or? 
image

mlocardpaulet:
> > > > @enryH I renamed the parameter test file with i2masschroq using xtandem, and added the one using sage (that we now need to parse). In the process, I noticed that this file `test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel.csv` (renamed`test/params/i2mproteobench_2pep_fdr01psm_fdr01prot_sel_xtandem.csv`) does not seem to be used. Or did I miss something? If not, shall we remove it?
> > > 
> > > 
> > > The test fail because it is missing. the `csv` version is always the version I want to see parsed ('truth'). I'll fix the tests now.
> > 
> > 
> > Thanks. I get it, sorry. And this one? `i2mq_result_parameters.tsv` it seems alone and unused. No?
> 
> I think it's an updated version of the file I am using in the test. Should we use both or either or? image

Hum... I am not sure. I see that the search engine version is not at the same location. This is why it is "NA" in the most recent public point of i2MassChroQ.
Maybe better to keep both.

proteobench/io/params/i2masschroq.py
```diff
@@ -73,27 +59,101 @@ def extract_params(fname: pathlib.Path) -> ProteoBenchParameters:
         fragment_mass_tolerance="[-" + _tol_frag + ", " + _tol_frag + "]",
         enzyme=_enzyme,
         allowed_miscleavages=max_cleavage,
-        min_peptide_length=None,  # "spectrum, minimum fragment mz"
-        max_peptide_length=None,  # Not mentioned, up to 38 AA in peptides
+        min_peptide_length=None,  # xtandem: "spectrum, minimum fragment mz"
```
enryH, line 62: So this would be 150, but that is quite indirect. Would you translate it by deviding by the maximum weight on an amino acid?

mlocardpaulet: I would not infer. 
For the max, why did you comment "up to 38 AA in peptides"? Where does this information come from? If it is reliable, why not using this? It seems quite short compared to other search engines so it may be interesting to have it parsed.

mlocardpaulet (commented):
It looks all good to me.

proteobench/io/params/i2masschroq.py
enryH: I think it was based on the available example I had.

mlocardpaulet: OK, so then do not put anything.
What we could consider is inferring min/max length from the result file. I'll discuss this in https://github.com/orgs/Proteobench/discussions/496

RobbinBouwmeester (approved):

PR #510, opened 2024-12-17
Peptidoform output from Proteome Discoverer

mlocardpaulet:
I added a test file in `test/data/dda_quant/ProteomeDIscoverer_PeptideGroups_trimmed.txt`. It is trimmed, and has all possible columns that can be exported from this level.

### DIFF
--- a/test/test_parse_params_i2masschroq.py
+++ b/test/test_parse_params_i2masschroq.py
@@ -9,7 +9,7 @@ TESTDATA_DIR = Path(__file__).parent / "params"
 
 
 def test_extract_params():
-    file = TESTDATA_DIR / "i2mproteobench_2pep_fdr01psm_fdr01prot.tsv"
+    file = TESTDATA_DIR / "i2mproteobench_2pep_fdr01psm_fdr01prot_xtandem.tsv"
     expected = pd.read_csv(TESTDATA_DIR / f"{file.stem}_sel.csv", index_col=0).squeeze("columns")
     actual = params_module.extract_params(file)
     actual = pd.Series(actual.__dict__)


mlocardpaulet:
From what I understand:
- there is one peptidoform per row
- stripped sequences are in `Sequence`
- modifications are in `Modifications`
- all protein accessions matching a given peptidoform and the species are in the column `Protein Accessions` (this requires a specific parsing rule of the fasta that I'll need to add in the documentation)
- multiple accessions are separated with ";"
- contaminants can be identified with "Cont_" in the column `Master Protein Descriptions`. **actually this may be an issue? I should maybe change the parsing rule so that it is in the column `Protein Accessions`. Or we need to have somewhere a list of all the contaminant accessions in the fasta.** Or maybe we can use the column `Contaminant`? But I would have to double check how this was set up in the search parameters.
- then for quant I would suggest to use the values of the columns:
Abundances Normalized   F1 Sample ConditionA, Abundances Normalized F2 Sample ConditionA, Abundances Normalized F3 Sample ConditionA, Abundances Normalized F4 Sample ConditionB, Abundances Normalized F5 Sample ConditionB, Abundances Normalized F6 Sample ConditionB
**the sample names may vary depending on how users set up their analysis. I have to check how to make it consistant...**

So... a few points are not completely clear yet, sorry.

mlocardpaulet:
Regarding the **contaminants**:
I think we should do as if the "Cont_" were in the column `Protein Accessions`. 
I will change the parsing of the fasta so that it works. 
It means that with the file that I uploaded right now, no contaminants will be removed but should work on the next one.

mlocardpaulet:
So... here is a new file, where I did not indicate anything in terms of experimental plan, so the column names where the quantities are are:

Abundances   (Normalized): F1: Sample
Abundances (Normalized): F2:   Sample
Abundances (Normalized): F3:   Sample
Abundances (Normalized): F4:   Sample
Abundances (Normalized): F5:   Sample
Abundances (Normalized): F6:   Sample

Sub-optimal. I'll have to find something else. I suspect that they are ordered (A, A, A and B, B, B), but I am not entirely sure.

But the issues with the column `Protein Accessions` are fixed: it now contains all the accessions matching a given proteoform AND the prefix "Cont_" for contaminants.
[MulticonsensusProteoBench_DDAmodule2_SequestHT_Percolator_Quanti_241218_PeptideGroups_trimmed.txt.zip](https://github.com/user-attachments/files/18183797/MulticonsensusProteoBench_DDAmodule2_SequestHT_Percolator_Quanti_241218_PeptideGroups_trimmed.txt.zip)

I will work on the documentation, and try to find a "simple" way to get column headers that make more sense for the quantification...

PR #528, opened 2025-01-10
add information on metrics calculation in docs - ion modules only

mlocardpaulet:


mlocardpaulet:
Hi @rodvrees, you were right. There was already a description in the documentation on how the metrics were calculated. I don't know why I missed/forgot this. 
So in the end I just rephrased it the same way for the two ion modules. Can you check that there is no mistake?
I also specified what columns were used for each tool in the documentation. I know that it is redundant with the information in the toml, but the non-informaticians need to know too. I did it only for the DDA module. If you think that what I did is OK, I'll also do it for the DIA. 
I tried to be clear when we filtered out data etc... But in the end, it is not with so many tools that we have to do weird stuff. 
Let me know what you think, and take your time. No hurry.

docs/available-modules/2-quant-lfq-ion-dda.md
```diff
@@ -64,10 +66,10 @@ The module is flexible in terms of what workflow the participants can run. Howev
...
 
-To submit your run for public usage, you need to upload the parameter file associated to your run in the field `Meta data for searches`. Currently, we accept outputs from MaxQuant, FragPipe, Proline Studio, and i2MassChroQ (see bellow for more tool-specific details). Please fill the `Comments for submission` if needed, and confirm that the metadata is correct (correspond to the benchmark run) before checking the button `I confirm that the metadata is correct`. Then the button 
+To submit your run for public usage, you need to upload the parameter file associated to your run in the field `Meta data for searches`. Currently, we accept outputs from MaxQuant, FragPipe, Proline Studio, and i2MassChroQ (see bellow for more tool-specific details). Please fill the `Comments for submission` if needed, and confirm that the metadata is correct (corresponds to the benchmark run) before checking the button `I confirm that the metadata is correct`. Then the button
```
rodvrees, line 69: There's more supported tools, right? Sage and Alphapept, off the top of my head

docs/available-modules/4-quant-lfq-ion-dia-aif.md
```diff
@@ -65,7 +67,7 @@ The module is flexible in terms of what workflow the participants can run. Howev
...
 
-To submit your run for public usage, you need to upload the parameter file associated to your run in the field `Meta data for searches`. Currently, we accept outputs from DIA-NN, AlphaDIA, FragPipe, MaxDIA and Spectronaut (see bellow for more tool-specific details). Please fill the `Comments for submission` if needed, and confirm that the metadata is correct (correspond to the benchmark run) before checking the button `I confirm that the metadata is correct`. Then the button 
+To submit your run for public usage, you need to upload the parameter file associated to your run in the field `Meta data for searches`. Currently, we accept outputs from DIA-NN, AlphaDIA, FragPipe, MaxDIA and Spectronaut (see bellow for more tool-specific details). Please fill the `Comments for submission` if needed, and confirm that the metadata is correct (corresponds to the benchmark run) before checking the button `I confirm that the metadata is correct`. Then the button
```
rodvrees, line 70: 1) Same applies here, we support more tools
2) Bellow -> Below

rodvrees (commented):
Thanks @mlocardpaulet, here's some comments.
The first few comments in the DDA docs also apply for the DIA docs :-)

mlocardpaulet:
Thanks guys, you are amazing. This was on my to-do list but was not quick enough.

Issue #520, opened 2025-01-06
Support for PEAKS in Proteobench (DDA and DIA)


brvpuyve:
Support for PEAKS Studio (DDA and DIA)

I have generated output for both the DDA and DIA module.

For DDA, there is both input for the ion (lfq.features.csv) and peptidoform (lfq.peptides.csv) level. 
The search parameters are exported as parameters.txt
https://cloud.cmb.ugent.be/index.php/s/zdGB3zZ7Fwed9gq?path=%2FModule_2_DDA_quantification%2Fsearch_results%2FPEAKS 

For DIA, there is also both input for the ion (lfq.features.csv) and peptidoform (lfq.peptides.csv) level. 
The search parameters are exported as parameters.txt
https://cloud.cmb.ugent.be/index.php/s/zdGB3zZ7Fwed9gq?path=%2FModule_4_DIA_Quantification%2FPEAKS

Unfortunately, PEAKS uses simplified sample naming which has resulted in the following:
Sample 1: A_REP1
Sample 2: A_REP2
Sample 3: A_REP3
Sample 4: B_REP1
Sample 5: B_REP2
Sample 6: B_REP3

- [x] Work on documentation 
- [x] Create toml file
- [x] Parse input file
- [x] Parse parameters for public

RobbinBouwmeester:
You are able to upload DDA ion, DDA peptidoform, and DIA AIF ion data with the code on this branch:

#522 

I am not sure if the chosen columns are correct, maybe you can check @brvpuyve:

https://github.com/Proteobench/ProteoBench/pull/522/files#diff-78dc5b6127d5bb0e589cd30146b493eae2d10bbead934d070d64b479486792a0
https://github.com/Proteobench/ProteoBench/pull/522/files#diff-16c0677be301c65e69b20900cdd0c48ebbb8752f2846a64f44fb303d50ffdfd3
https://github.com/Proteobench/ProteoBench/pull/522/files#diff-69eca82f8f769cffe6da085e63297b688dac9d5473da4055cf272345e2442c30

brvpuyve:
[...]

PR #534, opened 2025-01-17
drafingt parameter file parsing for peaks #520

wolski:


wolski:
image
@rodvrees, can you help? Locally, all tests are passing. I even clone proteobench again, but here in at GitHub the new test fails.

RobbinBouwmeester (approved):

PR #543, opened 2025-01-21
Msangel compatibility

mlocardpaulet:
Parameter parsing of MSAngel json files for Mascot searches. 
The parsing will be a bit different with other search engines (like X!Tandem - see file `test/params/MSAngel_Xtandem-export-param.json`). I am working on it!


mlocardpaulet:
also, I still have to make the python file for testing parameter parsing.

proteobench/io/params/__init__.py
```diff
@@ -89,6 +89,6 @@ class ProteoBenchParameters:
     max_precursor_charge: Optional[int] = None
     scan_window: Optional[int] = None  # DIA-specific
     quantification_method: Optional[str] = None
-    second_pass: Optional[bool] = None  # DIANN specific
+    second_pass: Optional[bool] = None  # used in both DDA and DIA: same thing?
```
enryH, line 92: This is new?

mlocardpaulet:
... And I just realised! I forgot to add the "hidden modifications of X!Tandem" `"quickPyrolidone" : true,`! 
I have to do it, but I am not entirely sure how yet.

proteobench/io/params/__init__.py
mlocardpaulet: well, yes. It is not "DIANN specific", so I changed the comment. 
I am sorry the new comment is not super clear. I just was not sure that it is exactly the same thing for dia and dda. But within each, I think that "second pass" is understandable. So it wont be an issue when we have module-specific parameters.

enryH (approved):
Code wise it looks good. But for now not too many parameters could be parsed looking at the extracted `csv` files.

mlocardpaulet:
> Code wise it looks good. But for now not too many parameters could be parsed looking at the extracted `csv` files.

I fully agree. I think that I can maybe add a few with David's help. But I am not sure how much we'll cover actually.

proteobench/io/params/__init__.py
RobbinBouwmeester: It is DIA specific though isnt it?

mlocardpaulet: well, we also say "two-pass search" (in MQ, X!Tandem, maybe others...) when we do two successive searches with different tolerances.

mlocardpaulet: in DDA

mlocardpaulet: and it is a good parameter to have (I thought we did) in DDA

RobbinBouwmeester:
I will merge #542 into this branch soon to resolve conflicts and update other issues.

test/params/mqpar1.5.3.30_MBR_sel.json
```diff
@@ -1,8 +1,8 @@
 {
-    "software_name": null,
+    "software_name": NaN,
```
enryH, line 2: null became NaN. But I think it would be good to put MaxQuant as a name anyways.

test/test_parse_params_maxquant.py
```diff
@@ -93,7 +93,10 @@ def test_flatten_of_dicts(dict_in, list_expected):
     assert actual == list_expected
 
 
-parameters = [(fname, fname.with_suffix(".csv")) for fname in mqpar_fnames]
+# TODO the test is broken, partly due to the expected files being incorrect
+# TODO skip for now, fix in future
```
enryH, line 97: you can make an issue and assign me... It was hacky from the start

test/test_parse_params_maxquant.py
```diff
@@ -103,10 +106,16 @@ def test_file_parsing_to_csv(file, csv_expected):
...
     actual = actual.to_frame("run_identifier")
     actual = pd.read_csv(io.StringIO(actual.to_csv()), index_col=[0, 1, 2, 3])
+    expected = expected.loc[actual.index]
+    print(actual)
+    print(expected)
     assert actual.equals(expected)
 
 
-parameters = [(fname, (fname.parent / (fname.stem + "_sel.json"))) for fname in mqpar_fnames]
+# TODO the test is broken, partly due to the expected files being incorrect
```
enryH, line 115: `null` changed to `NaN` - could be the issue (None vs NaN). but could be good to normalize that.

RobbinBouwmeester:
@enryH thank you for already looking, but I need to fix a lot here still ;)

RobbinBouwmeester:
image

Seems to be working with manual input!

RobbinBouwmeester:
Ok, everything seems to be working! Only thing that is left to do is indicate manually changed params in the PR.

RobbinBouwmeester:
Ok, it works! I am going to merge with main, we can do a little bit of testing and release a new version.

image

PR #544, opened 2025-01-21
Update analysis notebooks

rodvrees:


RobbinBouwmeester:
@rodvrees ready for review?

hollenstein:
Hi, I was wondering what is actually the purpose of the various jupyter notebooks? It doesn't seem to be documented anywhere. Or at least I couldn't find it. The only mentioning of the notebooks I've found is in the [CONTRIBUTING.md](https://github.com/Proteobench/ProteoBench/blob/main/CONTRIBUTING.md#running-tests-building-notebooks-and-documentation) file.

In general, jupyter notebooks are not really well suited for version control, but there are tools that convert a notebook into a python file (which works well for version control) and then back to a jupyter notebook when you want to run it...

rodvrees:
Hi, in general the ideas behind the ipynb's is to showcase to the end users how to use ProteoBench to perform analyses and functionality that go beyond what the web interface does. As of right now, these notebooks are mostly for me and Caro to share our code to create figures for the manuscript. They probably ended up in a PR quite prematurely, but the idea would be that we can simply include the ipynbs in the repo so that the manuscript analysis is fully open. That said, the version control element is definitely something we could discuss because you're right that they are not great for that, altough the notebooks serve more as examples, so normally there shouldn't be too much changes to them after they are merged to main. At least that's what I think the idea is.

hollenstein:
Thanks Robbe for the explanation.

Besides the analysis folder, there is also a server resubmission Notebook and a "submission" folder, containing also a server resubmission notebook and two others. Then there is also some kind of test notebook ...

Is there any good reason why all of these notebooks need to stay with the proteobench code? Have you considerd moving the notebooks to a separate repository, like 'proteobench-notebooks'. I believe having a dedicated repository could have several advantages. Here are just some ideas that you can think about =)

**Better documentation**
- You could use the notebook repositorys readme to clearly explain what the purpose of the different notebooks are, and how to use them. So if someone from outside (not a contributer) lands on the notebooks repository, it's immediatly clear what the purpose of the notebooks are.
- It's easier to make a clear (and understandable) separation between what the purpose of the notebooks is, because the repository would only contain the notebooks and not lots of other stuff.
	- e.g. some notebooks might be more "educational" for developers to showcase how proteobench works, there could be a dedicated folder to reproduce the manuscript figures, and there could be notebooks that help you with in-depth inspection of the benchmark results.

**Version control**
- Notebooks could be added as .py files to make version control and pull-request reviews much easier. There are tools that make this very easy (automatic conversion from .ipynb to .py and vice versa, also automatic execution of .ipynb).
- It would propably even be possible to make a pre-commit hook that keeps the .py files in sync with the .ipynb files, so you can still continue working solely in notebook format.

**Requirements**
- The notebooks might have additional requirements compared to the main proteobench repository, such as jupyter or ipykernel, or plotting libraries, ... So this repository could have its own pyproject toml file to make installation and using the notebooks easier.

**Clear separation of concern**
- The purpose of the proteobench main repository is to contain the code for the streamlit page and the underlying logic of the proteobench modules.
- The notebooks serve a different purpose such as  allowing inspection and additional analysis of the data provided by proteobench.

Issue #560, opened 2025-01-27
Param/input file parsing mistakes in currently public benchmarking runs


rodvrees:
I manually checked all currently public benchmark runs to identify some mistakes in parameter or input proforma parsing.
Safe to say there's some mistakes to iron out. Some probably need a little discussion.
Here's the rundown:

### **DDA ion quant**
**General**

- (minor) homogenize modifications

**Maxquant**

- Turns out, if you change PSM FDR in the GUI it changes the 'peptideFDR' setting in the mqpar.xml file. This value is parsed as fdr_level_peptide in ProteoBench, but should maybe be parsed as fdr_level_psm. There doesnt seem to be a peptide FDR level setting in the GUI.
- Fixed modifications are not parsed because they are not explicitely reported in MaxQuant (#545) 
- MaxQuant_20241216_130203 is not available on the server
- MaxQuant_20241216_134651 is not available on the server

**ProlineStudio**

- N-terminal modification parsing is incorrect, "-" is missing
Right now: [Acetyl]ACDEFGHI
Should be: [Acetyl]-ACDEFGHI
**_EDIT: this error already fixed, but point needs to be resubmitted to reflect fix (#535)_**
- (very minor) It would be easier if the output file could be downloaded as xlsx rather than txt because the txt won't be readable until extension is changed

**i2MassChroQ**

- Correctness of cleavage pattern -> protease needs to be checked by someone who is familiar with the tool, because it is unintuitive to check by someone who doesn't know the tool
- Minimum charge is hardcoded as 1 but no Charge 1 IDs were made, so is this correct?
- N-terminal modification parsing is incorrect
Right now: A[Acetyl]CDEFGHI
Should be: [Acetyl]-ACDEFGHI
**_EDIT: Issue is that i2MassChroq proforma column is directly used without parsing. Indeed in the i2MC output the acetylation is also put on the first AA instead of N-term. Is this intentional behaviour? Probably need to ask devs._** 
- In the runs with Sage as search engine, Oxidation is incorrectly parsed as mod:00719. This is not an issue with the X!Tandem runs
**_EDIT: #561, will also no longer be an issue with newer versions of iMass2ChroQ, as per https://github.com/Proteobench/ProteoBench/pull/561#issuecomment-2621773529_** 

**AlphaPept**

- According to the codebase of Alphapept, peptide_FDR parameter is used to do precursor level scoring, so should be parsed as psm_fdr? (ask devs)
[...]

PR #561, opened 2025-01-28
Fix Oxidation parsing i22MassChroQ + Sage

rodvrees:
As mentioned in #560


rodvrees:
Hi Olivier, 

For general oxidation, MOD:00425 makes the most sense indeed.
MOD:00719 is specifically for methionine oxidation, so essentially it is not 'incorrect', but personally I don't think it makes sense for PSM annotation to make the modification amino acid-specific.

Thanks for checking it out!

OlivierLangella:
Yes indeed, I was using MOD:00719 because the oxidation on methionine was asked for in the search engine.
But it's OK to be more generic and use MOD:00425.

This is fixed in i2MassChroQ :
https://forgemia.inra.fr/pappso/i2masschroq/-/commit/d597c6f78a29dc82f70720484eea9adad79c9640
ready for the next release.

Thanks
Olivier

RobbinBouwmeester:
Will still merge so older versions will also be correctly parsed.

RobbinBouwmeester (approved):

PR #575, opened 2025-02-10
chore: Remove files not intended for repo root

hollenstein:
Remove two files from the repo root directory that most likely ended up there by accident:

**test_proline.csv**
- This file seems to be generated by a test, but mabye isn't cleaned up afterwards

**searchindex.js**
- This file seems to belong into the built documentation here `\docs\_build\html\searchindex.js`, not sure how it could have ended up in the repo root? Or was this intentional?


enryH (commented):
Yes indeed. Seems to have happened accidentally when refactoring.

rodvrees (approved):

PR #583, opened 2025-02-15
:memo: update affiliation

enryH:
update my affiliation

### DIFF
--- a/docs/general-information/5-contributors.rst
+++ b/docs/general-information/5-contributors.rst
@@ -120,7 +120,7 @@ People who contributed to ProteoBench, in alphabetical order:
 
 .. line-block::
     **Henry Webel**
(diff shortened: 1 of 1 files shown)


RobbinBouwmeester:
You are allowed to bypass rules and merge if you want to @enryH  :P

enryH:
okay:) I will merge small ones, but I think for larger ones it's nice to have some reviews.

mlocardpaulet:
Or @enryH  you could put me in charge of reviewing the affiliations. 
And... Are you sure that you don't want to keep both? Since you also worked on ProteoBench while being in your previous position?

enryH:
yes that is true, but do you normally keep old affiliations? For the article you could add both, yes.

mlocardpaulet:
You are right. We should keep current affiliation on the website and all affiliations in the paper :)

PR #590, opened 2025-03-05
Comply docstrings with numpydoc, add numpydoc-validation to pre-commit hooks

rodvrees:
This pull request, which is in preparation of merging the subcell module, is primarily focused on improving documentation and code comments across the package, i.e., converting method docstrings to NumPyDoc style, and updating pre-commit hooks.

You will notice a load of files are "changed", including figures. These are just EOF "fixes" (removing or adding empty line(s)) to comply with the pre-commit hook "fix end of files". Sorry that this makes this PR annoying to review.

We should probably discuss how rigidly we should enforce pre-commit hooks (and which) for development and GitHub actioning in a meeting


rodvrees:
Alright that  py3.8 test fail was annoying, perhaps a good idea to get working on #576

hollenstein:
Looks good. Thanks for correcting the mistakes. I've also checked the images, they still look the same. Not sure why it needs to change line endings in image files, but as long as it doesn't harm...

Yeah, let's at least drop 3.8 support =)

hollenstein:
Looks good. Thanks for correcting the mistakes. I've also checked the images, they still look the same. Not sure why it needs to change line endings in image files, but as long as it doesn't harm...

Yeah, let's at least drop 3.8 support =)

hollenstein (commented):
Looks good. Thanks for correcting the mistakes. I've also checked the images, they still look the same. Not sure why it needs to change line endings in image files, but as long as it doesn't harm...

Yeah, let's at least drop 3.8 support =)

hollenstein (approved):
Sorry, forgot to select approve ...

RobbinBouwmeester (approved):

PR #592, opened 2025-03-05
Update supported Python versions

hollenstein:
Related to the discussion about dropping support for old Python versions: #576 

I've created this pull request to get the discussion going. So far it includes the following changes:

- Drop support for Python version 3.8, 3.9, 3.10
- Add support for version 3.12 and 3.13
- Set default Python version to 3.11

I've chosen the version 3.11 as the default one because it is used in the `automated_release` workflow. But I actually don't really have an opinion if the default version should be 3.11 or a newer one.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.11", "3.12", "3.13"]
 
     steps:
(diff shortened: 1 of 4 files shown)


julianu:
Fine for me. With 3.12 we would have another year of support and it should be decently stable. But no strong opinion, as long as it is above 3.10.

julianu (approved):

hollenstein:
A agree with Julian about 3.12, is anyone against also dropping support for 3.11 and directly moving to 3.12?

Cajac102 (approved):

rodvrees (approved):

PR #613, opened 2025-04-01
single-cell module

SamvPy:
The PR includes an implementation of a new single-cell module as discussed in #426 and is largely based on the other DIA LFQ ion benchmarking modules.


proteobench/plotting/plot_quant.py
```diff
@@ -32,11 +32,11 @@ def plot_fold_change_histogram(result_df: pd.DataFrame, species_ratio: Dict[str,
         go.Figure
             A Plotly figure object representing the histogram.
         """
+        SPECIES = [species for species in species_ratio.keys()]
```
RobbinBouwmeester, line 35: Is this a global variable? Otherwise this should probably be lowercase.

RobbinBouwmeester (approved):

PR #738, opened 2025-05-24
Update submission tab layout

rodvrees:
Updates submission tab layout to:

1) Not show printed json and thus only submission form
2) Only show free-text comment field and check box once parameter file has been uploaded
3) Move checkbox to bottom


webinterface/pages/base_pages/quant.py
```diff
@@ -1109,17 +1110,19 @@ def display_public_submission_ui(self) -> None:
         """
         Display the public submission section of the page.
         """
+        # Initialize Unchecked submissoin box variable
+        st.session_state[self.variables_quant.check_submission] = False
```
rodvrees, line 1114: Had to do this to initialize the key in the session state, otherwise gave errors. But not sure if this is the way to approach it?

PR #740, opened 2025-05-24
Cfg flag param parsing

rodvrees:
So I learned that the cfg file is just a freetext file that contains any part of the command to run DIA-NN. 
Current approach works, but it ain't winning any beauty contests.
We could consider asking userss not to use --cfg flags in the docs, I'm not sure it's often used


proteobench/io/params/diann.py
```diff
@@ -51,6 +88,8 @@
 ]
 SETTINGS_PB_MOD = ["fixed_mods", "variable_mods"]
 
+PROT_INF_MAP = {"isoform IDs": "Isoforms", "protein names": "Protein_names", "genes": "Genes"}
```
RobbinBouwmeester, line 91: Should the lines just be lowered? (mainly talking about the "genes" : "Genes"  part.

rodvrees: No, its to conform the parameters when free text or command line is parsed

RobbinBouwmeester (approved):

PR #745, opened 2025-05-26
update the module life cycle doc

mlocardpaulet:
I updated the module life cycle document to correspond to our discussions last week. 
Not sure it is all good. 
Feel free to amend/comment.

### DIFF
--- a/docs/general-information/2-module-life-cycle.md
+++ b/docs/general-information/2-module-life-cycle.md
@@ -3,7 +3,7 @@
 Anybody can propose a new benchmark module, and discuss on the validity of current modules. There are 6 phases in the benchmark modules life cycle:
 
 1. Module proposal
-2. Implementation
+2. Alpha (late phase of implementation)
 3. Beta 
 4. Live
 5. Archived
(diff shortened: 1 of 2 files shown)


rodvrees:
Does this imply that if the beta banner is removed, there can be no added tools, or param parsing fixes?

enryH:
I would say the approval is rather for the conceptual benchmark proposed in the module, not the implementation of parameter and tool output parsing.

mlocardpaulet:
I guess that the beta banner and the stamp of external expert approval (although the point is more to have the discussion, I would not use "approval") are two independent things. 
But I find it very difficult to clearly define when to remove the beta banner. We said that it would be at time of publication for the modules that we are currently working on. But if we get useful reviewer(s) comments, we may change modules quite a bit...

mlocardpaulet:
Maybe we should start a discussion on when to remove the beta banner? 😅

rodvrees:
I agree with @enryH that the BETA flag should not really reflect the implementation of param or tool parsing, since indeed there might always be errors we want to fix and eternal BETA flags are indeed not desirable.

If we want to keep the BETA banner and external expert approval banner separate, I do like the idea of the beta banner being a time-dependent thing, like you write in the docs. That there is a certain amount of time needed that is used to collect feedback and bug reports. Altough this is still a bit vague, of course

PR #751, opened 2025-05-26
make file for documenting what parameters are parsed for public submi…

mlocardpaulet:
…ssion


rodvrees:
I think it's a good idea to have, because people can refer to it if we make errors toward param parsing, instead of having to look at the code or spot wrongly parsed parameters in their submissions. I think people will find it useful. I'm willing to help out with writing the tool-specific param parsing, if you want help. How to make sure that it stays up to date is a big question. I think small changes, at least in the beginning, will occur quite regularly. It could be automized I guess, if we wanted to. Just some kind of mapping of original setting parsed in the relevant file. But I'm not sure how to exactly combine this with readthedocs (I don't have that much experience with it), and it might be overkill

docs/available-modules/12-parsed-parameters-for-public-submission.md
```diff
@@ -2,20 +2,20 @@
...
 
-## General comments on the parameters that are retreived by ProteoBench
+## General comments on the parameters that are retrieved by ProteoBench
 
 The parameters that are retrieve from the parameter files are the following:
 
 - "software_name": name of the software tool.
 - "software_version": version of the software tool used to generate the run.
-- "search_engine": name of the search engine used for MS-based identification.
+- "search_engine": name of the search engine used for MS-based identification. Often the same as the software name, but is relevant for data analysis workflows that allow multiple search engines to be used (TODO: example)
```
enryH, line 11:
```suggestion
- "search_engine": name of the search engine used for MS-based identification. Often the same as the software name, but is relevant for data analysis workflows that allow multiple search engines to be used (e.g. quantms)
```

enryH:
So the dataclasses are built using now a specification for parameters used in modules, stored as json in [proteobench/io/params/json/Quant](https://github.com/Proteobench/ProteoBench/blob/HEAD/proteobench/io/params/json/Quant)

I can add some @rodvrees, e.g. for quantms if you want. but I guess you go through the parser and update the origin? 

I hope if we start this, we can keep it in sync...

rodvrees:
> So the dataclasses are built using now a specification for parameters used in modules, stored as json in [proteobench/io/params/json/Quant](https://github.com/Proteobench/ProteoBench/blob/HEAD/proteobench/io/params/json/Quant)
> 
> I can add some @rodvrees, e.g. for quantms if you want. but I guess you go through the parser and update the origin?
> 
> I hope if we start this, we can keep it in sync...

Sorry, I didn't see your comment 😬. I think I did what I wanted to do, feel free to change whatever you want  @enryH, @mlocardpaulet. 

Pretty sure, right now, all jsons in [proteobench/io/params/json/Quant](https://github.com/Proteobench/ProteoBench/blob/HEAD/proteobench/io/params/json/Quant) are identical. But I guess as soon as we're making very different modules, we will want different params to be parsed. 
How we make sure this stays updated is still a big question. Maybe my approach is not very good for that (if you want to completely change it, feel free), but on the other hand, the only way I can imagine this being automized is having an easy key=value mapping of parameter=parsed value, but the parameter parsing itself is not simplistic enough for that to be feasible, IMO...

mlocardpaulet:
I cannot approve this PR because I created it 😆.
But I approve. Nice work.

enryH:
I will still add the `parsing_overview.py` to the build process on the docs.

enryH:
So it was actually linked in the docs on the parameters overview page: https://proteobench--751.org.readthedocs.build/en/751/available-modules/12-parsed-parameters-for-public-submission/

For now we use everywhere the same json for parameters as @rodvrees noticed and he is implementing something to fix that.

I opted for testing in the tests that the `parsing_overview.tsv` is unchanged, prompting for manuel update as you can see in the failed action for testing https://github.com/Proteobench/ProteoBench/actions/runs/16219823735/job/45797454801

RobbinBouwmeester (approved):

Issue #681, opened 2025-05-21
make documentation link more visible in the app


mlocardpaulet:
people miss the sentence "The full description of the module is available [here](https://proteobench.readthedocs.io/en/latest/available-modules/5-quant-lfq-ion-dia-Astral/)" that is on top of all module pages. 
We need to make it more visible, maybe with an icon, a banner, a button? 

I can work on it. Happy to get suggestions.

PR #755, opened 2025-05-27
update list of modules

mlocardpaulet:


mlocardpaulet:
While I am at it, I will have a look at information buttons to make the documentation page more visible on the webpage (see https://github.com/Proteobench/ProteoBench/issues/681)

PR #810, opened 2025-07-04
Pmultiqc

enryH:
- add pmultiqc reports to ProteoBench DDA precursor quant module
- need to compute report based on intermediate file (either using a subprocess or directly in code)

HTML report can be downloaded and then opened in the browser directly.

To discuss:

- Should the report be always generated? 
  - then maybe saved to the temporary folder?
  - if not: compute only on request as it takes a bit of time?
  
Before merging:

- [ ] could be added to a tab so that it is reflected in all reports (currently on separate tab only for the first module (dda quant)

### DIFF
--- a/webinterface/pages/2_Quant_LFQ_DDA_ion_QExactive.py
+++ b/webinterface/pages/2_Quant_LFQ_DDA_ion_QExactive.py
@@ -1,6 +1,7 @@
 """
 Streamlit UI for the DDA quantification - precursor ions module.
 """
+
 import logging
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


RobbinBouwmeester:
@enryH should we already review/merge this?

enryH:
Yes I would go ahead. Pmultiqc is ready and it would be great to have it integrated. I will have to test it again with the merged main branch. 

Is there a way to create the reports for all submitted data on the server myself? I guess only @julianu can run scripts on the it?

- [ ] allow people to do download the pmultiqc report for already submitted data (tab1)

enryH:
With the latest `pmultiqc` version and the changes in #862 it does currently not work.

pyproject.toml
```diff
@@ -34,7 +34,8 @@ dependencies = [
   "PyGithub",
   "GitPython",
   "psm-utils",
-  "python-calamine"
+  "python-calamine",
+  "pmultiqc>=0.0.29",
```
RobbinBouwmeester, line 38: With the intermediate format changes this should be >v0.0.33?

webinterface/pages/2_Quant_LFQ_DDA_ion_QExactive.py
```diff
@@ -122,6 +126,50 @@ def _main_page(self) -> None:
...
+                tmp = st.session_state[self.quant_uiobjects.variables_quant.result_perf]
+                tmp_path = Path('tmp_pmultiqc')
+                tmp_path.mkdir(parents=True, exist_ok=True)
+                tmp.to_csv(tmp_path / 'result_performance.csv', index=False)
+                subprocess.run(
+                    ["multiqc", "--parse_proteobench", "./tmp_pmultiqc", "-o", "./report"],
+                    check=True,
+                )
+
+                st.write("Intermedate data saved to: ", tmp_path)
```
RobbinBouwmeester, line 147: Is this for debugging?

RobbinBouwmeester (changes requested):

webinterface/pages/2_Quant_LFQ_DDA_ion_QExactive.py
enryH: I will remove it and log it instead. I need to move the code also in a separate module so it can be shared between all quant benchmarking modules.

PR #836, opened 2025-07-28
Parse min max mz for DIA modules

rodvrees:
Introduces mz parsing for the DIA modules.
Also, completes the logic which allows different params to be parsed in a module-specific way using the jsons in ProteoBench/proteobench/io/params/json/Quant

### DIFF
--- a/proteobench/io/params/alphadia.py
+++ b/proteobench/io/params/alphadia.py
@@ -76,7 +76,7 @@ def parse_key_value(line: str) -> Tuple[str, str]:
 
 def detect_newer_version(lines: List[str]) -> bool:
     """
-    Detect if the log file is from a newer version of AlphaDIA based on the presence of 'user defined' in the lines.
+    Detect if the log file is from a newer version (>= 1.10) of AlphaDIA based on the presence of 'user defined' in the lines.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


proteobench/io/params/alphadia.py
```diff
@@ -69,9 +74,31 @@ def parse_key_value(line: str) -> Tuple[str, str]:
...
+    """
+    for line in lines:
+        cleaned_line = clean_line(line)
+        # If user defined and default are in the same line, it indicates a newer version
+        if "user defined, default" in cleaned_line:
+            return True
+    return False
+
+
+def _extract_values_newer_version(lines: List[str], start_index: int) -> List[int]:
```
RobbinBouwmeester, line 99: Would be good to document the version that is considered "newer"

proteobench/io/params/alphadia.py
```diff
@@ -85,7 +112,36 @@ def extract_values_from_nested_lines(lines: List[str], start_index: int) -> List
...
+        if cleaned_line:
+            number = re.search(r"\d+", cleaned_line)
+            if number:
+                values.append(int(number.group(0)))
+        if len(values) == 3:  # Assuming we want three values
+            break
+    return values[:-1]  # Exclude the last value which is often not needed
+
+
+def _extract_values_older_version(lines: List[str], start_index: int) -> List[int]:
```
RobbinBouwmeester, line 127: Would be good to document the version that is considered "older"

RobbinBouwmeester (approved):

PR #839, opened 2025-07-30
Fix i2mq parsing wihtout enzyme restriction field

rodvrees:
Also reimplements the module-specific parameter parsing, which got lost after the UI refactoring


proteobench/io/params/i2masschroq.py
```diff
@@ -102,6 +102,8 @@ def _extract_sage_params(
     ProteoBenchParameters
         The extracted parameters encapsulated in a `ProteoBenchParameters` object.
     """
+    for param in params.index:
+        print(f"Parameter: {param}, Value: {params[param]}")
```
RobbinBouwmeester: This can probably be removed?

RobbinBouwmeester (approved):

PR #859, opened 2025-08-11
add datapoint comparison feature

Cajac102:
Vibecoded a feature for the comparison of two or more datapoints. Highlights parameters that are changed between two entries.

### DIFF
--- a/webinterface/pages/base_pages/tab1_results.py
+++ b/webinterface/pages/base_pages/tab1_results.py
@@ -7,7 +7,7 @@ import pandas as pd
 import re
 
 from proteobench.plotting.plot_quant import PlotDataPoint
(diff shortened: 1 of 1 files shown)


rodvrees:
Is there a way to combine the "selected" and "compare" columns?

Expected behaviour would be like this:
User clicks on a point in the plot => "selected" checkbox is ticked => added to compare set
OR
User ticks the "selected" checkbox => added to compare set

Currently I feel like having both "selected" and "compare" clutters the dataframe a bit with a lot of emptiness

Cajac102:
That would indeed be a cleaner solution. Would also enable comparing points completely without the table.
For now, I (=ChatGPT) struggle with the reverse connection of changing highlights in the plot according to a selection. I'll table it for now and check when I have more time to dig into it

Issue #697, opened 2025-05-22
Dataset, RAW and FASTA download


julianu:
To make the download of all datasets easier, it might be better to separate teh datasets from FASTA and RAWs.
For this, I propose to leave datasets in the "./datasets/" directory and move FASTA, RAW (and other stuff) to something else... maybe "./data/" or less confusing use two folders "./fasta/" and "./raws/"?

mlocardpaulet:
this looks good. I totally agree. 

When you do that, could you get the data for the DIA-PASEF module and put it on the server? You can find them here: https://www.ebi.ac.uk/pride/archive/projects/PXD062685

PR #860, opened 2025-08-12
Move location of raw files and FASTA on the server

julianu:
The RAWs and FASTA file are moved on the server. This PR reflects this in the documentation.

The FASTA is now at https://proteobench.cubimed.rub.de/fasta/
The RAWs are now at https://proteobench.cubimed.rub.de/raws/
Processed datasets are still at https://proteobench.cubimed.rub.de/datasets/


julianu:
This implements #697

douweschulte:
A quick notification (very likely you are already working on this) but this means that the docs should be updated to reflect the new locations. Notably https://proteobench.readthedocs.io/en/latest/available-modules/active-modules/2-quant-lfq-ion-dda/#how-to-use contains a link to the fasta file that is broken now but worked after removing `datasets/` from the url. The same link is present here: https://proteobench.readthedocs.io/en/latest/available-modules/active-modules/5-quant-lfq-ion-dia-diapasef/ and here https://proteobench.readthedocs.io/en/latest/available-modules/active-modules/7-quant-lfq-precursor-dia-Astral_2Th/ and here: https://proteobench.readthedocs.io/en/latest/available-modules/active-modules/8-quant-lfq-precursor-dda-Astral/.

mlocardpaulet:
This looks very good to me, thanks. Just one comment: the folder names may be a bit too generic ("DIA" and "DDA"). It does not have to be done now, but I wonder if it would be a good idea to have a tree-based structure similar to what was done for the modules?

julianu:
Yes, a structure based more on the modules would also be nice. I will keep this in mind when we will make the larger work on the server (re-submissions etc.)
Though some modules also share the same RAWs right now (DDA peptidoform and ion)

PR #862, opened 2025-08-12
change charge notation to adhere proforma rules

### LINKED ISSUE #861: Use ProForma syntax for denoting charges
Currently the charges of peptidoforms in the output data looks as follows: `[Acetyl]-TEQM[Oxidation]TLR|Z=2`. The charge addendum is a nonstandard addition to the otherwise valid ProForma sequence. In ProForma 2.0 there is an extension to allow charges `SEQUENCE/<charge>`. In the upcoming version 2.1 of ProForma this is moved into the main specification, but the notation has stayed exactly the same. With this change the peptide would be `[Acetyl]-TEQM[Oxidation]TLR/2`.

Having this changed would have made it easier for me to add a parser as part of my [project](https://github.com/rusteomic/rustyms) to parse all manner of identified peptidoform files.

Cajac102:
Fixes #861 

Go from |Z=charge to /charge to adhere to proforma 2.0 rules

Until points are resubmitted, this will lead to slightly inconsistent intermediate results, however this is pretty easily fixable locally (in excel as well)

### DIFF
--- a/proteobench/io/parsing/parse_settings.py
+++ b/proteobench/io/parsing/parse_settings.py
@@ -352,7 +352,7 @@ class ParseSettingsQuant:
         """
         if self.analysis_level == "ion":
             if "proforma" in df.columns and "Charge" in df.columns:
-                df["precursor ion"] = df["proforma"] + "|Z=" + df["Charge"].astype(str)
+                df["precursor ion"] = df["proforma"] + "/" + df["Charge"].astype(str)
             return df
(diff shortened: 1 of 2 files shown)


rodvrees (approved):

enryH:
Can you document the changes in an example. So the new format uses a `/` to separate the sequence from the charge? `AAAAAAALQAK/2` ? And before it was a `|` ?

PR #914, opened 2025-11-12
Module banners update

mlocardpaulet:
Hi, 
the main aim was to modify the banner of the archived module (it was still "beta"). 
In the end I reorganised the banner text and conditions. Could you check?
I add the ShortMessages of the generic text document the messages for the banners (alpha, beta and archived). 
And I modified the module pages accordingly.
Could you please confirm that you are OK with this new system for the banners? I'll work on some documentation for the Alpha. Put it somewhere. 

Question: 
Why are there 2 documents for the generic texts? One for DIA and one for DDA? Are these both used?


rodvrees (approved):

docs/general-information/2-module-life-cycle.md
```diff
@@ -27,9 +27,13 @@ After two positive code reviews by ProteoBench maintainers, the PR can be merged
...
 
+## Alpha
+
+When the PR is merged, the new module enters an alpha stage, where its code base is part of the Python package, and it is present on the web platforms. However, a prominent banner states that the module is still in "Alpha". This means that the module is available but remains to be fully discussed with experts (more information on module proposal process can be found [here](./3-module-proposal.md)). 
+
 ## Beta
 
-When the PR is merged, the new module enters a beta stage, where its code base is part of the Python package, and it is present on the web platforms. However, a prominent banner states that the module is still in “Beta”. After a minimal period of one month and approval by the initial proposers and external reviewers, the beta label can be removed.
+The new module enters a beta stage when it has been discussed and approved by the community.  Its code base is part of the Python package, and it is present on the web platforms. However, a prominent banner states that the module is still in “Beta”. After a minimal period of one month and approval by the initial proposers and external reviewers, the beta label can be removed.
```
enryH:
```suggestion
The new module enters a beta stage when it has been discussed and approved by the community. Its code base is part of the Python package, and it is present on the web platforms. However, a prominent banner states that the module is still in “Beta”. After a minimal period of one month and approval by the initial proposers and external reviewers, the beta label can be removed.
```

mlocardpaulet:
> I would say the message is repeated too often in the same page (with 4 tabs). Can you consider to create a function for it, so it can be changed in one place only?

Sure, well, at least now the message in the banner is only on one location.
How would you see it?

webinterface/pages/10_Quant_LFQ_DIA_ion_ZenoTOF.py
```diff
@@ -102,15 +107,20 @@ def _main_page(self) -> None:
...
-                    icon="🚨",
+            if self.variables_dia_quant.archived_warning:
+                st.info(
+                    self.variables_dia_quant.texts.ShortMessages.warning_archived
                 )
-            if self.variables_dia_quant.beta_warning:
-                st.warning(
-                    "This module is in BETA phase. The figure presented below and the metrics calculation may change in the near future."
+            else:
+                if self.variables_dia_quant.alpha_warning:
```
enryH: this is repeated from above. I am not sure how you want the warnings exactly to display, but I think this can go into an function, checking ing if the configuration sets these flags and then it calls the display method.

enryH: yes. I will quickly have a look.

mlocardpaulet:
All good on my side!

rodvrees:
I think @enryH needs to accept the requested changes and then we can merge.

PR #928, opened 2025-11-28
Add Species weighted metrics logic

### LINKED ISSUE #924: Add species-weighted metrics for HYE quant modules so that quant accuracy metric is not dominated by most abundant human precursors
### Discussed in https://github.com/orgs/Proteobench/discussions/923

<div type='discussions-op-text'>

<sup>Originally posted by **rodvrees** November 28, 2025</sup>
In the current modules, the median and mean log2FCs are calculated globally, i.e. across all precursors. Because in the current HYE samples from the Van Puyvelde et al. dataset (data from which is used for almost all current modules), human precursors are vastly more represented in the sample mixes, the median and mean log2FCs will be predominantly dominated by the error on human precursors, and less by others. This is why it is important to also have metrics that are weighted equally for each species, because, if anything, the quant accuracy on the lower abundant species' precursors is more important. 

The user ideally should be able to choose whether to visualize the overall mean/median log2FC, or a species-weighted metric.
Work on the implementation on this is already underway (actually it's finished), but implementing this on the already existing modules requires the recalculation (i.e. resubmission) of all currently existing datapoints, which is not something we have automated, at the moment.
Once we have automated resubmission of existing datapoints, we can implement the new toggle.</div>

rodvrees:
Introduces new metrics where mean/median log2FCs is now not only calculated globally across all precursors, but also separately for each species and then averaged (equally weighted species). See discussion #923 for reason why.

Since current datapoints do not contain these metrics, backward compatibility is ensured by treating the original metrics and the new "global" equally (they are, the name was just changed), and only showing the datapoints that contain the equally weighted species metrics in the plot if that mode is selected by the user (currently there are none, so users will see a descriptive warning message)

Once we have automatic submission, we can recalculate the new metrics on existing datapoints, and then we can clean up some of the backward compatibility-ensuring code, which causes some complexity. I also suggest, that once we have done this, we select the equal weighted species metric as the default

closes #924, at least the implementation, but obviously #925 is needed to ensure that all points can be visualized using the new  metrics.

### DIFF
--- a/webinterface/pages/base_pages/tab1_results.py
+++ b/webinterface/pages/base_pages/tab1_results.py
@@ -214,7 +214,7 @@ def display_metric_calc_approach_selector(variables_quant) -> str:
     return st.radio(
         "Select metric calculation approach",
         options=["Global", "Species-weighted"],
-        help="Toggle between species-weighted and global absolute difference metrics.",
+        help="Toggle between species-weighted and global absolute difference metrics. Global considers all metrics equally, while species-weighted account for species abundance variations, i.e. the mean/median metric is calculated per species first before averaging across species.",
         key=_id_of_key,
     )
 


rodvrees:
Looks good to me right now. The ROC-AUC metrics are currently not shown in the UI, but we can make that happen if and when we want, and do that in a different PR. @RobbinBouwmeester, can you review, then we can merge.

webinterface/pages/base_pages/tab4_display_results_submitted.py
```diff
@@ -134,47 +140,30 @@ def initialize_submitted_data_points(
         )
 
 
-def handle_submitted_table_edits(variables_quant) -> None:
```
RobbinBouwmeester: Can this be left out? Was this moved?

PR #960, opened 2026-02-19
Multi-species protein group module

mlocardpaulet:
Hey, this is very much a test. Not sure about my coding but I wanted to try. 
It is not done, for sure.

Also, I made changes to the quantscore to make it flexible (not only for precursors). I am not sure it is the best idea.


mlocardpaulet:
I would very much like comments etc.... 
- I know that the datapoints are not perfect (for example there are still mentions of precursors when it should be protein groups). 
- we have to discuss metrics (what to plot on the y axis). 
- we want to add specific calculations: number of unique accession per group for example, and one to check if there is parcimony (number of duplicated accessions across groups)
- I only added DIA-NN outputs (pg_matrix)
- I think that the contaminants are not filtered out, for this we should use the Protein.Group identifiers

mlocardpaulet:
During the in-person meeting, we decided that for clean comparison, and to know what expected ratio to match to each protein group, I will only consider the protein accessions that have no shared peptide in theory (according to the fasta file). And we will only report the accuracy error for this. With the number of accessions on the y-axis.

mlocardpaulet:
TOFIX: `default_cutoff_min_prec` and `default_cutoff_min_feature`

Issue #1025, opened 2026-04-17
Switch to Custom Naviagation


enryH:
Instead of defining `pages_variables.py` files which define which sides are rendered, we could switch the explicit creation of a site navigation:

https://docs.streamlit.io/develop/concepts/multipage-apps/page-and-navigation#adding-section-headers

This would mean that the filtering function might not be there as it is now, but it could be added as a search field which just list pages.

See: https://docs.streamlit.io/develop/concepts/multipage-apps/page-and-navigation#adding-section-headers

@rodvrees @wolski @RobbinBouwmeester : Any comments why this could not work?

PR #1026, opened 2026-04-17
Debug session state

enryH:
Page to debug session state. 
- Table of variables with type and their repr (abrev. for lists and dataframes)

Adding the page was not intuitive and I would consider updating the building of the app to be more expressive. See #1025


enryH:
@rodvrees I do not see per default the sidebar. So the sidebar has to be build for each page separately? I cannot see this in any of the module pages explicitly. I had to add it via

```python
import pages.texts.proteobench_builder as pbb
pbb.proteobench_sidebar(current_page="/")
```

which I find a bit confusing.

rodvrees:
@enryH 

The sidebar is initialized within each module page through BaseUIModule constructor in pages/base.py:

```
class BaseUIModule(ABC):
    """Base class for all UI modules with common patterns."""
     def __init__(self, variables, ionmodule, parsesettingsbuilder, page_name: str = "/") -> None:
        ...
        pbb.proteobench_sidebar(current_page=self.page_name)
```

And each of the module pages just initializes a BaseUIModule

RobbinBouwmeester:
@enryH ready for review?

RobbinBouwmeester:
Merge for now, @enryH can reopen if needed

PR #1028, opened 2026-04-17
change prec to features in variables names such as nr_prec

### LINKED ISSUE #1019: change "nr_prec" to "nr_feature"
since we use the same functions for different modules at precursor, protein group, peptidoform level... We should change "nr_prec" to "nr_feature" in the code.

mlocardpaulet:
I made the replacements, seems to work. 
In the table below the main plot with public points the columns is still "nr_prec" for the precursor module and the peptidoform module -> TO FIX.
Also, the y-axis labels should match the module ("peptidoform count" for the peptidoform module).


mlocardpaulet:
OK, looks good but now everything is referred to as "number of features". Maybe we want to specify, at least in the axis titles, when it is number of precursor ions or peptidoforms etc...

mlocardpaulet:
Now fix the tests.

mlocardpaulet:
@enryH could you check?
We had "nr_prec" (i.e. number of precursors) everywhere, even in modules where we calculated peptidoform counts. It bothered me. Also, since we want to add a protein group module, I wanted to change "nr_prec" to "nr_feature".
(see https://github.com/Proteobench/ProteoBench/issues/1019)

So, I think that I managed to change all occurrences, with back-compatibility. 
In the tables below the main plots, the column is now labelled "nr_feature". Hope this is fine. 

I think that it does not change anything regarding the plasma and de novo module (@rodvrees @SamvPy). 

I had to add a variable y_axis_title in the _variable.py pages.
I changed the test file test_quan_datapoint.py.

Constructive criticisms are welcome. Hope I did not make it more complicated than needed.

enryH:
Hey. Changes in names are always quite tedious. You searched for all occurrences of `prec`? 

The change in the tables might break others downstream analysis scripts if they rely on column names. I hope they manage to change that.

enryH:
what about `prec_df`?

mlocardpaulet:
> Hey. Changes in names are always quite tedious. You searched for all occurrences of `prec`?
> 
> The change in the tables might break others downstream analysis scripts if they rely on column names. I hope they manage to change that.

I see your point. Then would you be in favor of not changing it? 
Wouldn't it be confusing in the long run?

I also see now that there is this `precursor_column_name` that we should consider changing.

RobbinBouwmeester (approved):

PR #1029, opened 2026-04-21
Handle no species names in Proteins output MaxQuant

rodvrees:
We should probably investigate what causes MaxQuant output not to have species info in Proteins column (only accession).
Might be MaxDIA specific. Anyway, this handles both.


Cajac102 (approved):
We should figure out when is this actually the case, seems strange that it sometimes happens. Maybe another fasta was used?

rodvrees:
It was not the fasta but a Parse setting probably, but this should circumvent taht

PR #1032, opened 2026-04-27
Fix regex for multiple days in timestamp alphadia

rodvrees:
This pull request enhances the parameter extraction logic for AlphaDIA logs and expands test coverage to handle edge cases involving unusual peptide length values. The main changes include improving the timestamp parsing regex, adding a new test input file, and introducing a corresponding test case.

**Improvements to parameter extraction:**

* Updated the `TIMESTAMP_REGEX` in `alphadia.py` to more robustly handle timestamps with optional "days"

**Test coverage enhancements:**

* Added a new test input file `alphadia_weird_lengths.csv` with edge-case parameter values, particularly targeting unusual peptide length settings.
* Included the new test file in the list of files processed by the `extract_params` test, ensuring that the code is validated against these edge cases.


proteobench/io/params/alphadia.py
```diff
@@ -13,7 +13,7 @@
 
 # Regular expression to clean up lines from ANSI escape codes
 ANSI_REGEX = re.compile(r"(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]")
-TIMESTAMP_REGEX = re.compile(r"(\d+ day,)|(\d+):\d{2}:\d{2}\.?\d*")
```
RobbinBouwmeester, line 16: Is this for a specific version? Was this always the case?

rodvrees: Always the case, we just never had runs that ran for more than one day before :-D

PR #1052, opened 2026-06-02
Parameter validation

### LINKED ISSUE #993: Checks and balances
Check if input parameter file matches search results (charge, peptide length...)

Check fasta matches the ones we approved (hash, easter egg protein, ...)

### LINKED ISSUE #1022: PR from Proteobot: set up a system for the moderators

RobbinBouwmeester:


RobbinBouwmeester:
Should implement #993

proteobench/validation/checks.py
```diff
@@ -0,0 +1,1063 @@
...
+    "argc": ("R", False),
+    "gluc": ("DE", True),
+    "chymotrypsin": ("FYWL", False),
+    "lysn": None,
+    "aspn": None,
+}
+
+#: Plausibility ceilings for mass-tolerance sanity checks.
+_MAX_PLAUSIBLE_PPM = 1000.0
+_MAX_PLAUSIBLE_DALTON = 10.0
```
rodvrees: Defaullt sage config is "precursor_tol": {
        "da": [
            -500,
            100
        ]

so mass tolerant searches will be flagged by default?

proteobench/validation/checks.py
```diff
@@ -0,0 +1,1063 @@
...
+
+    Returns
+    -------
+    str
+        The normalized enzyme key.
+    """
+    text = str(name).strip().lower()
+    for char in (" ", "_", "-"):
+        text = text.replace(char, "")
+    return text
```
rodvrees, line 106: We should maybe wait for #990 then, since for example DIA-NN and metamorpheus do not report enzyme like this but rather as cleavage sites

rodvrees: When #990 is merged this should be no problem anymore

proteobench/validation/fasta.py
```diff
@@ -0,0 +1,331 @@
...
+        for name in names:
+            if os.path.basename(name) == member_filename or name == member_filename:
+                return name
+    for name in names:
+        if name.lower().endswith(_FASTA_EXTENSIONS):
+            return name
+    raise ValueError(f"No FASTA member found in archive (members: {names}).")
+
+
+def parse_fasta_header(header: str) -> Set[str]:
```
rodvrees, line 97: Is the case handled where some tools output UP accession and other output gene name?

rodvrees: Yes it is

CLAUDE.md
```diff
@@ -403,6 +477,7 @@ Separate workflow for the webinterface (`test-streamlit.yml`):
 ### Adding a new benchmark module
 
 1. Create a module class in `modules/quant/` inheriting `QuantModule`, setting `module_id`, the precursor column, and repo names
+   - For submission validation (see the Submission Validation Layer section): add a `[reference_database] fasta_url = "..."` entry to the new module's `module_settings.toml` (if absent, protein-identifier validation is silently skipped). A quant module auto-resolves to the `quant_lfq` profile; a module of a new category should declare `[validation] profile = "<name>"` and register a matching profile in `proteobench/validation/profiles.py`.
```
rodvrees, line 483: The developer guide should also be updated with this information

rodvrees (changes requested):

proteobench/validation/checks.py
RobbinBouwmeester: Yes, but since the parameters always need to be submitted this should be retrieved from the parameter file. We could choose to not set these to default values, requiring them to be set by the parameter file.

proteobench/validation/checks.py
RobbinBouwmeester: As this is not a complete block to submission I would argue we can go ahead and this will be completely fixed once #990 is in there

CLAUDE.md
RobbinBouwmeester: Will do!

proteobench/validation/checks.py
rodvrees: I think that would be better, yes

PR #1099, opened 2026-07-09
Entrapment update

Cajac102:
-Scatter first, forest second
-important bug fix in FDP calculation
-get rid of annoying deprecated width calls (this has been a general issue for a while, I was finally annoyed enough to fix it :) )
-add support for FragPipe (DIANN) and AlphaDIA by Robbe
-add documentation for those by Robbe
-add documentation page to overview in readthedocs

Still missing:
Currently in discussion if we allow variable modifications or not. waiting on feedback.


proteobench/io/parsing/parse_settings.py
```diff
@@ -871,7 +871,6 @@ def __init__(self, parse_settings: Dict[str, Any], parse_settings_module: Dict[s
         self.mapper = parse_settings["mapper"]
         self.run_mapper = parse_settings["run_mapper"]
         self.decoy_flag = parse_settings["general"]["decoy_flag"]
-        self.contaminant_flag = parse_settings["general"]["contaminant_flag"]
```
rodvrees, line 874: was this unused?

rodvrees (approved):

proteobench/io/parsing/parse_settings.py
Cajac102: only in tests... maybe better to leave it in anyways

Cajac102: the tests were for other Parsers, for entrapment it is really never used. i dont discriminate between contaminant and organism IDs, they are all defined as targets. ill leave it out.

PR #1125, opened 2026-08-03
add custom format support for entrapment

Cajac102:
add support for custom file format for entrapment.
removed protein column from intermediate as it is not necessary for this module


proteobench/io/parsing/parse_ion.py
```diff
@@ -30,13 +30,11 @@ def load_input_file(input_csv: str, input_format: str, input_csv_secondary: str
     """
     try:
         if input_format == "MaxQuant":
-            warnings.warn(
-                """
+            warnings.warn("""
                 WARNING: MaxQuant proforma parsing does not take into account fixed modifications\n
```
rodvrees, line 35: probably causing linting error

rodvrees (approved):

PR #1135, opened 2026-08-06
Denovo module updates

SamvPy:
## Summary

This PR addresses feedback from two external reviewers on the *de novo*
sequencing (DDA-HCD) module, received during community review. Each
reviewer's comment is quoted or paraphrased below, followed by how it was
addressed — including cases where a deliberate decision was made *not*
to change something, for transparency.

---

### 1. Average precision (AUC)
**Comment:** Reporting average precision (area under the precision-coverage
curve) is more informative than precision at 100% coverage alone.

- [x] Added AUC as a metric option. Removed the "recall" toggle (see 3.
  below for why) and added a second tab on the in-depth plot showing full
  precision-coverage curves.

### 2. Amino acid substitutions (I/L, Q/E)
**Comment:** Most evaluations allow I/L substitution; uncertain whether
Q/E should also be allowed — worth showing both ways.

- [x] Added a toggle to allow/disallow I/L ambiguity and a separate toggle
  for Q/E (deamidation) ambiguity, so results can be viewed either way.

### 3. Recall is not well-defined for de novo sequencing
**Comment:** The notion of a false negative (required for a standard
recall definition) doesn't apply to de novo sequencing, since there's no
true negative class.

- [x] Removed "recall" from the module. Replaced with **coverage**
  (proportion of spectra for which a prediction was made) and **AUC**
  (area under the precision-coverage curve), which don't rely on an
  undefined false-negative concept.

### 4. Database search vs. matching precision
**Comment:** Worth evaluating precision by matching predicted peptides
(≥8 amino acids) against a reference proteome, as a complement to the
standard database-search gold standard.

- [x] Added this as a new in-depth plot: a stacked bar chart per tool
  showing the proportion of predictions that are **correct**,
  **incorrect but present in the reference proteome ("in FASTA")**, or
  **incorrect and not found in the proteome**.

---

### 1. Main benchmarking plot is hard to interpret
**Comment:** Plotting amino-acid-level and peptide-level metrics against
each other may obscure interpretation, since different tools show
different AA-vs-peptide performance profiles with different
implications. Suggested separate swarmplots, or at minimum fixing the
axis ranges to [0, 1].

- [x] Kept the combined scatterplot (rather than switching to
  swarmplots), but fixed both axis ranges to [0, 1] and added
  quadrant-based color coding with explicit labels describing what a
  tool's position in each corner implies. Applied the same treatment to
  the PTM-related scatterplots.

### 2. Recall vs. coverage
**Comment:** Coverage is a clearer concept than recall for de novo
sequencing tools (see linked reference).

- [x] Same fix as Reviewer 1, item 3 above — recall replaced with
  coverage and AUC.

### 3. I/L should not be treated as equivalent in exact matching
**Comment:** Isoleucine and leucine should not automatically be
considered equivalent in exact-match evaluation, given evidence some
tools can distinguish them.

- [x] Addressed via the same I/L toggle described in Reviewer 1, item 2 —
  equivalence is now optional rather than assumed.

### 4. Terminology inconsistency (tooltip vs. docs)
**Comment:** The plot menu says "Mass-based" while the tooltip says
"Match-based"; mass tolerance units are described inconsistently
(Da vs. ppm) across different doc sections.

- [x] Fixed. Terminology and units are now consistent between the UI,
  tooltips, and documentation.

### 5. Benchmark Results table columns
**Comment:** Current columns seem tool-specific and may not be relevant
for all submissions; worth clarifying which are expected to be
standardized vs. tool-dependent.

- [ ] **Not changed in this PR.** Columns are kept as-is for now. Open to
  revisiting in a follow-up once we've gathered feedback from more tool
  submissions on which columns are actually useful across the board.

---

# Closing remark

Finally, I'd like to thank both reviewers for their contributions! Feel free to comment further and we'll gladly accomodate/discuss any points related to the module

### DIFF
--- a/docs/available-modules/active-modules/11-denovo-dda-hcd.md
+++ b/docs/available-modules/active-modules/11-denovo-dda-hcd.md
@@ -44,32 +44,35 @@ We recommend downloading the parsed and combined dataset from the ProteoBench se
 
 ## Metric calculation
 
-The performance is evaluated at both the amino acid and peptide level. As introduced by [DeepNovo](https://www.pnas.org/doi/10.1073/pnas.1705691114), a correct amino acid whose mass differs by less than 0.1 Da from the corresponding ground truth amino acid. Additionally, this predicted amino acid must have either a prefix or suffix that differs by no more than 0.5 Da in mass from the corresponding amino acid sequence in the ground truth peptide. Correct peptides are defined as sequences where all amino acid predictions meet these criteria, ensuring that only fully accurate predictions are considered correct at the peptide level. In the module, this mode of evaluation is called '**mass-based**'. However, a more strict evaluation mode can be selected and is termed '**exact mode**'. In this mode, the two sequences should be exactly the same, where also cases such as deamidated-Q and E are considered incorrect. Only isoleucine and leucine substitutions are allowed.
+The performance is evaluated at both the amino acid and peptide level. As introduced by [DeepNovo](https://www.pnas.org/doi/10.1073/pnas.1705691114), a correct amino acid whose mass differs by less than 0.1 Da from the corresponding ground truth amino acid. Additionally, this predicted amino acid must have either a prefix or suffix that differs by no more than 0.5 Da in mass from the corresponding amino acid sequence in the ground truth peptide. Correct peptides are defined as sequences where all amino acid predictions meet these criteria, ensuring that only fully accurate predictions are considered correct at the peptide level. In the module, this mode of evaluation is called '**mass-based**'. However, a more strict evaluation mode can be selected and is termed '**exact mode**'. In this mode, the two sequences should be exactly the same. However, you can specify to allow mistakes made between isoleucine-leucine and deamidated-DE - NQ.
 
 ### Main benchmarking plot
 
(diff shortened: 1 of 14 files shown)


Cajac102 (commented):
Could you add a description and/or example custom format in the documentation?

If I change the plot options, the plots turn empty.
If I select Exact, the plots are empty. mass based and precision works, AUC and mass based is empty. I guess there is a bug, or do the points need to be resubmitted to calculate the new metrics?

Under "How are the metrics calculated", it is still explaining recall 

The FASTA Analysis plot in the Compare Results tab is also empty, even though points are selected

rodvrees:
Screenshot 2026-08-06 151640

This is quite a dark background, can we make it a little more opaque?

Screenshot 2026-08-06 151747
Precision-Coverages curves seem to be empty right now, is this because points have not been resubmitted?

Screenshot 2026-08-06 151830
Needs to be updated. This tab is new, so this PR is older than it. You can find the metric explanation in plot_generator_denovo.get_metrics_help_markdown and plot_generator_denovo.get_in_depth_metrics_help_markdown

SamvPy:
@Cajac102 
1. Description of custom format was [here](https://proteobench.readthedocs.io/en/stable/available-modules/active-modules/11-denovo-dda-hcd/#custom-format) in documentation but fixed a bug there to optionally provide amino acid level confidence scores.

2. Plots turn empty for new features, because the old submitted datapoints don't have the proper format/elements in there. If you try to upload a new datapoints, the plots should work.

3. Fixed the 'metrics calculation' description

4. Is explained similarly as point 2.

@rodvrees 
1. Increased opacity
2. See point 2 above
3. See point 3 above

Oh and I added support for PEAKS output

rodvrees (approved):

PR #1145, opened 2026-08-15
Add InstaNovo v1.2.2 denovo parsing

BioGeek:
This PR makes the parsing of InstaNovo configs more robust. It now supports parsing InstaNovo v1.2.2 config files and handles config files from older InstaNovo versions. It now also reads precursor mass tolerance, which was not extracted at all previously.


SamvPy (approved):
Thank you for the contributions! Tested upload of results file and configuration file, checked changed docs, and test run without errors.

All looks good to me. Ready to merge

PR #1147, opened 2026-08-15
Give generated datapoint IDs sub-second resolution

BioGeek:
`generate_id()` stamps the ID with whole-second resolution:

```python
time_stamp = datetime.now().strftime("%Y%m%d_%H%M%S")
self.id = "_".join([self.software_name, str(time_stamp)])
```

so two datapoints created within the same second get the same ID.

`generate_datapoint()` already builds a microsecond-precision ID in the constructor:

```python
id=input_format + "_" + user_input["software_version"] + "_" + formatted_datetime,   # %Y%m%d_%H%M%S_%f
...
result_datapoint.generate_id()                                                        # overwrites it
```

and then `generate_id()` runs immediately afterwards and overwrites it, so the coarser value is the one that survives.

You will never hit this when submitting by hand. But datapoints are also generated in batches: `resubmit_datapoints.py` regenerates many in a loop, where several per second is normal.

A duplicate ID is not a data-loss risk, results are stored as `{intermediate_hash}.json`, and that hash is derived from the data itself. But the ID is what the results table displays, what row highlighting matches on (`df["id"].apply(lambda x: "➡️" if x == highlight_id else "")` would mark both rows), and what the submission branch name is built from.

## The change

One line per class: `%Y%m%d_%H%M%S` → `%Y%m%d_%H%M%S_%f`, plus a docstring line saying why.

Applied to all three datapoint classes — `QuantDatapointHYE`, `DenovoDatapoint`, `EntrapmentDatapoint`, which carry identical copies of `generate_id()`.

The ID shape is otherwise unchanged (`<software>_<timestamp>`), and only newly generated IDs are affected; stored datapoints keep theirs.

### DIFF
--- a/proteobench/datapoint/denovo_datapoint.py
+++ b/proteobench/datapoint/denovo_datapoint.py
@@ -197,9 +197,12 @@ class DenovoDatapoint(DatapointBase):
         """
         Generate a unique ID for the benchmark run by combining the software name and a timestamp.
 
-        This ID is used to uniquely identify each run of the benchmark.
+        This ID is used to uniquely identify each run of the benchmark. The timestamp carries
+        microseconds because datapoints are also generated in batches -- the resubmission
+        script regenerates many in a loop -- where whole-second resolution hands several runs
+        the same ID.
         """
-        time_stamp = datetime.now().strftime("%Y%m%d_%H%M%S")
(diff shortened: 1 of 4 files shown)


rodvrees (approved):
LGTM! Thanks!

PR #1136, opened 2026-08-07
Add module in discussion for FDR validation in documentation main page

mlocardpaulet:

PR #1149, opened 2026-09-04
Fix formatting and add discussion entry Xlink

mlocardpaulet:


mlocardpaulet:
Already done in https://github.com/Proteobench/ProteoBench/pull/1136

PR #1150, opened 2026-09-14
Replace create_distplot, removed in plotly 7

BioGeek:
## Why CI is red everywhere

plotly **7.0.0** was released on **2026-08-25** and removed `plotly.figure_factory.create_distplot`. `pyproject.toml` pins `plotly>=6.0.0`, so every CI run since then resolves to 7.0.0 and fails collection on six test modules:

```
ImportError: cannot import name 'create_distplot' from 'plotly.figure_factory'
```

`main` last ran green on **2026-08-21**, four days before that release, so this is not a regression from any pull request — it simply lands on whichever one runs next. It is currently failing #1146, and `main` itself would fail if re-run.

## The fix

Both call sites (`plot_generator_lfq_HYE.py`, `plot_generator_lfq_PYE.py`) passed `show_hist=False, show_rug=False`, so the only thing the figure factory contributed was the kernel density estimate. `proteobench/plotting/utils.py` now builds that directly with `scipy.stats.gaussian_kde` — scipy is already a core dependency, so nothing is added.

**The replacement reproduces the old figure exactly, not approximately.** Same 500-point grid spanning each group's own min to max, same default bandwidth, same trace names, legendgroups, marker colours and layout. Checked by building both figures from identical input under plotly 6.2.0:

```
traces 3 == 3,  max|dx| = 0,  max|dy| = 0,  styling fields identical
```

Pinning `plotly<7` was the alternative. This seemed better: the pin would have to come off eventually, and the usage here was a thin slice of the figure factory rather than anything hard to reproduce.

## One deliberate behaviour change

Groups with fewer than two distinct values are now skipped rather than raising. `gaussian_kde` inverts the sample covariance, so a single point or a constant run is singular — previously one degenerate species would have taken down the entire plot.

## Verification

Built a clean environment the way CI does (`pip install .[web]`, resolving plotly 7.0.0) and ran the suite:

| | before | after |
|---|---|---|
| collection errors | 6 | **0** |
| tests passed | 0 (interrupted) | **573** |

Also confirmed the new helper still behaves identically on plotly 6.2.0, so the `>=6.0.0` pin stays honest.

Four `TestEntrapmentFDPCalculation` tests are deselected in my local run: they download over HTTPS and this machine cannot verify the certificate. They fail the same way without this change, and they passed on CI in the 2026-08-21 run that introduced them.

`black --check`, `flake8 --select=E9,F63,F7,F82` and the `numpydoc-validation` hook all pass on the changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PBVfTpgipQhBg7ZqjkcDo8


rodvrees:
LGTM, thanks!

rodvrees (approved):

